PR #3, opened 2023-05-05
simplify

rafaqz:
These are from Turf.jl but very heavily refactored.
 
I realised I had already half converted turf to the latest geointerface :rofl:


rafaqz:
Sure! I haven't used the package myself but it looks good, although maybe not fast as it requires duplication of all objects as tuples. Many algorithms here will not allocate.

In this PR I am just hacking in someone elses code from Turf.jl (which copied turf.js) with a focus on getting the high level algorithms working on any GeoInterface.jl compatible objects in a clean, generic way.

The lower level predicates can really be swapped out at any point, feel free to jump in and hook them up if you want to.


I've actually move the predicataes out to another PR for now too.

rafaqz:
@DanielVandH  now that I look at it, I cant see what predicates from ExactPredicates.jl would replace `is_point_on_segment`. What would you suggest?

DanielVandH:
Maybe since this is something that'll be considered properly down the line, I'll put some of my thoughts about this in an issue and respond there.

rafaqz:
@skygering @asinghvi17 as I mentioned in #8, I've written up `simplify` with variants for the major algorithms, selectable with constructors:

```julia
simplify(RadialDistance(; tol=0.1), obj)
simplify(DouglasPeucker(; tol=0.01), obj)
simplify(VisvalingamWhyatt(; number=7), obj)
```

And the default  uses `DouglasPeucker` as its the most common:
```julia
simplify(obj; tol=0.1)
```

What do you think about this as a general pattern?

Also, `simplify` is built on `apply` so it will simplify over any object. There is way of selecting a `Union` of types with `apply` that seems pretty useful when you need to have different algorithms for different geometry traits:
https://github.com/asinghvi17/GeometryOps.jl/blob/5aad1e958434ed93e302ec6250cb26d922178cae/src/transformations/simplify.jl#L59

PR #13, opened 2023-05-29
add polygonize

rafaqz:
Generates polygons from a matrix.

Still needs winding detection to separate exteriors and holes


asinghvi17:
Hey, is the polygoize function only supposed to distinguish one from zero?  Or should we allow a list of levels as kwargs (perhaps only for the Contours algorithm representation)?

rafaqz:
Good question. I guess matching some condition like `x in 2.1..4.7` or `x == 3` is identical in practice?

So maybe they can all do that.

asinghvi17:
I just took a look at GDAL and it looks like `GDALPolygonize` emits polygons for each unique value in the array.  

We could either do that, and emit polygons with metadata somehow embedded (or return `(polygons, values)`), or say that these functions only work on boolean matrices and move forward with that.  What do you think?

rafaqz:
We could emit Feature objects in a feature collection/Table

rafaqz:
Lets merge this for now so we can use it, I will add more things later

PR #19, opened 2023-09-28
Make centroid use GeoInterface

skygering:
This is not done. I still need to write a few more tests / figure out why the multi polygon test is failing. However, I wanted to get feedback as I am still getting the hang of GeoInterface.

Thanks!


rafaqz:
This looks really good to me, I didnt see anything I would change.

skygering:
Okay so I added more tests! I also added code I had for a random polygon generator to make it easier for us to create polygons as we go, or even just call the generator in our tests. No guarantees that it doesn't make self-intersecting polygons though (although I changed some things to make it less likely).

The other big change is that I switched from providing signed area to just providing area. It was too complex, especially with multi-polygons (what does signed area even mean for those?). I also had to take the absolute value of the area for calculating polygons and multi-polygons anyways to get the weighting correct.

skygering:
I also wanted to decide what the output should be. @rafaqz and @asinghvi17  do you envision these as giving back tuples, GeoInterface objects, or a geometry object equal to the type put in? We should try to be consistent across all of the functions.

I also want to get rid of the Float64, but not sure how. Any suggestions?

rafaqz:
Do you want to use the point type to set the float type?

You could do :

```julia
T = typeof(GI.x(getpoint(geom, 1)))
```

Then `zero(T)` is probably more correct than `T(0)`.


If you want the type to be equal yo the original type you could do

```julia
GI.convert(typof(GI.getpoint(geom, 1)), point)
```
But its not really guaranteed to work because e.g. Shapefile.jl has no GI.convert methods because they're not usefull objects except when reading a shapefile.

So I dont know what object to return really, it doesnt matter a lot at this stage.

skygering:
Okay great! I will switch to `zero(T)` and I think I will just have it return a Tuple then. It seems weird to return a GeoInterface Point when that might not be something that the user wants to deal with. So returning something that is a standard Julia type/object might be better.

Once I make those changes, am I good to merge?

PR #22, opened 2023-10-04
Sg/update intersects

skygering:
I updated and tested the `intersects.jl` file and also some other files that were connected got a few touch ups.

The details of the implementation are detailed at the top of `intersects.jl`.

Something I am not sure about is the use of the `meets` keyword in the `intersects` functions. I get how it works for a line segment (just two points) in that it can exclude endpoints. However, what does that mean for a line string (multiple segments) with a line through the segment that happens to pass through a vertex. Right now, `meets = 0` means that that isn't tagged as intersecting, when in my mind it is. However, this is a degenerate case. I guess I am not totally sure of the use cases for `meets`. Thoughts? 

I also changed a test case for the same polygon `overlap`ing with itself. This was false before, but I think it should be true as they are completely overlapping.


test/methods/bools.jl
```diff
@@ -128,7 +128,7 @@ import GeometryOps as GO
         (-53.34136962890625, 28.430052892335723),
         (-53.57208251953125, 28.287451910503744),
     ]])
-	@test GO.overlaps(pl3, pl4) == false
+	@test GO.overlaps(pl3, pl4) == true # this was false before... why?
```
skygering, line 131: This is the test that I changed!

rafaqz: No idea why it was wrong, I copied these from Turf.jl. I guess testing these against LibGEOS is best?

rafaqz:
Do we need `meets`? LibGEOS doesn't have that.

You can do `touches(a, b)` for meets?

test/methods/bools.jl
skygering: That makes sense! They were using intersects as equivalent to overlaps, which isn't valid I think since two of the same geometries on top of one another would have an intersection, namely the whole geometry. I will double check with LibGEOs though.

src/methods/bools.jl
```diff
@@ -193,6 +171,26 @@ function point_on_line(point, line; ignore_end_vertices::Bool=false)::Bool
     return false
 end
 
+function point_on_seg(point, start, stop)
```
skygering, line 174: I added this because the `point_on_segment` function was complicated with all of the potential endpoint options. Do we want all of those options? I removed the similar option `meets` as we discussed since it wasn't something that is available in LibGEOS. I can see how this might be helpful in other functions though...

src/methods/equals.jl
```diff
@@ -0,0 +1,192 @@
...
+    equals(::T, l1, ::T, l2) where {T<:GI.AbstractCurveTrait} ::Bool
+
+Two curves are equal if they share the same set of points going around the
+curve. 
+"""
+function equals(::T, l1, ::T, l2) where {T<:GI.AbstractCurveTrait}
+    # Check line lengths match
+    n1 = GI.npoint(l1)
+    n2 = GI.npoint(l2)
+    # TODO: do we need to account for repeated last point??
```
skygering, line 128: If there are two polygons, one with a repeated first point at the end, and one without the repeated point, should they still be equivalent?

test/methods/equals.jl
```diff
@@ -0,0 +1,104 @@
...
+    r1 = LG.LinearRing([[0.0, 0.0], [5.0, 5.0], [10.0, 0.0], [5.0, -5.0], [0.0, 0.0]])
+    r2 = LG.LinearRing([[3.0, 0.0], [8.0, 5.0], [13.0, 0.0], [8.0, -5.0], [3.0, 0.0]])
+    l3 = LG.LineString([[3.0, 0.0], [8.0, 5.0], [13.0, 0.0], [8.0, -5.0], [3.0, 0.0]])
+    # Equal rings
+    @test GO.equals(r1, r1)
+    @test GO.equals(r2, r2)
+    # Different rings
+    @test !GO.equals(r1, r2) && !GO.equals(r2, r1)
+    # Equal linear ring and line string
+    @test !GO.equals(r2, l3) # TODO: should these be equal?
```
skygering, line 38: If we have two curves, but one is a linear ring and one is a line string , but they have all the same points, should they be the same? I don't think so, but maybe...

test/methods/overlaps.jl
```diff
@@ -0,0 +1,105 @@
...
+
+    p3 = LG.Polygon([[[1.0, 1.0], [1.0, 6.0], [6.0, 6.0], [6.0, 1.0], [1.0, 1.0]]])
+    # Test basic polygons that overlap
+    @test GO.overlaps(p1, p3) == LG.overlaps(p1, p3)
+
+    p4 = LG.Polygon([[[20.0, 5.0], [20.0, 10.0], [18.0, 10.0], [18.0, 5.0], [20.0, 5.0]]])
+    # Test one polygon within the other
+    @test GO.overlaps(p2, p4) == GO.overlaps(p4, p2) == LG.overlaps(p2, p4)
+
+    # @test_throws MethodError GO.overlaps(pl1, (1, 1))  # I think these should be false
```
skygering, line 79: Is there something wrong with allowing this? This works for my code right now... should I find a way to make it throw an error?

rafaqz: Can a point overlap a popygon? The dimensions dont match

https://en.m.wikipedia.org/wiki/DE-9IM

rafaqz:
The passsing tests are on 1.9 too.

Looks like somwhere a Generator isnt being collected and that fails on 1.6 because `keys` isnt defined for ut. Maybe in `equals` or ` overlaps` ? ` getgeom` returns a Generator, it could be related to that...

test/methods/overlaps.jl
rafaqz: Oh right.

If a method here works for `Point` it should work for all `p` where `GI.trait(p) isa GI.PointTrait`

skygering:
Okay I just removed the call to `findfirst` and basically just wrote the code for `findfirst`. At least the tests all pass now.

test/methods/equals.jl
skygering: I have determined that these should be equal given LibGEOS outputs. Similarly, a multi-polygon that holds a single polygon is equal to that same polygon. So it is about what the shape contains, not about the shape type, although obviously, some types preclude equality.

src/methods/equals.jl
skygering: I think they should be. A polygon is closed by definition, so that is just a matter of representation if the last point is repeated.

skygering: Thoughts on this?

skygering: I updated this! It now works. Hopefully, my implementation isn't too complicated. I will push soon.

rafaqz: Sorry forgot to reple. Its harder for GeoInterface because Geome6ryBasics uses Linestring for everything so ww can enforce these things.

Glad you found a solution

skygering:
Is everyone okay with me merging?

rafaqz:
Thanks!! code looks great.

Issue #23, opened 2023-10-26
Optimizations


skygering:
Functions that could be parallelized (please add as we go so that we can optimize later):

- Signed area could conceivably be multithreaded

rafaqz:
Probably parallelising `apply` at the level of Vectors, FeatureCollections or MultiPolygons would be most effective. We could have a `threaded` keyword and make it false after we thread at the outer level.

asinghvi17:
https://www.crunchydata.com/blog/performance-improvements-in-geos may be a useful way to compare ourselves against LibGEOS (for at least some of the benchmarks)

PR #24, opened 2023-10-26
Thread for transformations like reproject, simplify, etc

rafaqz:
Add optional threading to everything that uses `apply`.

Threaded tasks will be used at the outermost level of the object passed in, over a vector of geometries, features in a feature collection, or e.g. polygons in a multipolygon.

See #23

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -24,16 +24,7 @@ flipped_geom = GO.apply(GI.PointTrait, geom) do p
     (GI.y(p), GI.x(p))
 end
 """
-function apply(f, ::Type{Target}, geom; calc_extent=nothing, kw...) where Target 
-    # Catch the type instability here in the outer method
-    # so false is nothing::Nothing and true is true::Bool
-    # we can think of a nicer way to do this later...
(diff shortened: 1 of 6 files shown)


skygering:
Why are we using a keyword for threaded rather than just making it multithreaded if there is more than one thread available? Is this common practice and will users expect this? Other than that, looks great!

rafaqz:
A few reasons.

We dont actually know how many threads are available any more, because threading can be nested.

If this is running in an already threaded context it may be better to avoid further threading.

But actually, the main reason I added the keyword is because our own code is nested (think a vector of feature collections of multipolygons) and we dont want to keep triggering threading at every level, just the outer one.

Then if we need the keyword anyway for internal use we may as well let users choose as well.

skygering:
That makes sense to me! Thanks for explaining. Happy to see it merged.

skygering (approved):

rafaqz:
No worries! I also might leave `false` as the default for now in case there are any threading related bugs. That could make debugging other things harder while we build things out.

PR #27, opened 2023-12-21
Sg/area dist

skygering:
Update and fully test the area and distance function.


src/methods/area.jl
```diff
@@ -0,0 +1,152 @@
...
+signed_area(geom) = signed_area(GI.trait(geom), geom)
+
+"""
+    area(::GI.PointTrait, point)::Real
+
+The area of a point is always zero. 
+"""
+function area(::GI.PointTrait, point)
+    T = typeof(GI.x(point))
+    return T(0)
```
rafaqz, line 67:
```suggestion
    return zero(T)
```

src/methods/area.jl
```diff
@@ -0,0 +1,152 @@
...
+signed_area(trait::GI.PointTrait, point) = area(trait, point)
+
+"""
+    area(::GI.AbstractCurveTrait, curve)::Real
+
+The area of a curve is always zero. 
+"""
+function area(::CT, curve) where CT <: GI.AbstractCurveTrait
+    T = typeof(GI.x(GI.getpoint(curve, 1)))
+    return T(0)
```
rafaqz, line 84:
```suggestion
    return zero(T)
```

rafaqz (commented):
Tests look great.

There are a lot of separate docs... I find it easier to read one doc string on a stub function covering all the methods of the function, rather than many separate per-method docs.

src/methods/area.jl
```diff
@@ -0,0 +1,152 @@
...
+
+Finds the area of a multi-polygon, which is the sum of the areas of all of the
+sub-polygons.
+"""
+area(::GI.MultiPolygonTrait, geom) =
+    sum((area(poly) for poly in GI.getpolygon(geom)))
+
+
+"""
+    _signed_area(geom)::Real
```
rafaqz, line 128: Underscore functions should have comments rather than docs

src/methods/area.jl
```diff
@@ -0,0 +1,152 @@
...
+really only makes sense for polygons, given with a multipolygon can have several
+polygons each with a different orientation and thus the absolute value of the
+signed area might not be the area. This is why signed area is only implemented
+for polygons.
+=#
+
+"""
+    area(geom)::Real
+
+Returns the area of the geometry.
```
rafaqz, line 49: Probably this function could have all of the `area` docs.

src/methods/area.jl
```diff
@@ -0,0 +1,152 @@
...
+"""
+area(geom) = area(GI.trait(geom), geom)
+
+"""
+    signed_area(geom)::Real
+
+Returns the signed area of the geometry, based on winding order.
```
rafaqz, line 56: And this one all of the `signed_area` docs

src/methods/area.jl
```diff
@@ -0,0 +1,152 @@
...
+    T = typeof(GI.x(p1))
+    area = T(0)
+    for i in 1:np
+        p2 = GI.getpoint(geom, i)
+        # Accumulate the area into `area`
+        area += GI.x(p1) * GI.y(p2) - GI.y(p1) * GI.x(p2)
+        p1 = p2
+    end
+    area /= 2
+    return area
```
rafaqz, line 151:
```suggestion
     return area / 2
```

skygering:
> Tests look great.
> 
> There are a lot of separate docs... I find it easier to read one doc string on a stub function covering all the methods of the function, rather than many separate per-method docs.

Okay! I combined all of the docs into the stub function and then organized the others with comments. Let me know if that is easier to read.

skygering:
Also, the code that was originally in the file had the `signed_distance` from a point to a multipolygon as the maximum minimum distance from a point to an element of the multipolygon. But based on LibGEOS behavior (as shown in tests) and this [description of Shapely behavior](https://gis.stackexchange.com/questions/459091/definition-of-multipolygon-distance-in-shapely), I think it should be the minimum minimum distance from the point to an element of the multipolygon. But let me know what you think.

rafaqz:
@asinghvi17 what do you think? I have no knowledge of signed distance algs to really comment on that

rafaqz:
The minimum minimum does seem correct to me too, based on not very much

skygering:
I feel good about merging it if you do @rafaqz. If someone disagrees with the algorithm we can always fix it later.

PR #29, opened 2023-12-22
actually test embed_extent

rafaqz:
I added this code without finishing or testing it. So this PR does that.

`embed_extent` embeds an `Extents.Extent` object in all geometries recursively. This means `extent` calls are basically free when you need them later.

Useful as preparation for algorithms that need to call `Extents.extent` on everything.

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -1,10 +1,19 @@
+const THREADED_KEYWORD = "- `threaded`: `true` or `false`. Whether to use multithreading. Defaults to `false`."
+const CRS_KEYWORD = "- `crs`: The CRS to attach to geometries. Defaults to `nothing`."
+const CALC_EXTENT_KEYWORD = "- `calc_extent`: `true` or `false`. Whether to calculate the extent. Defaults to `false`."
+
+const APPLY_KEYWORDS = """
+$THREADED_KEYWORD
+$CRS_KEYWORD
+$CALC_EXTENT_KEYWORD
+"""
 
 # # Primitive functions
 
 # This file mainly defines the [`apply`](@ref) function.
 
 """
-    apply(f, target::Type{<:AbstractTrait}, obj; crs)
+    apply(f, target::Type{<:AbstractTrait}, obj; kw...)
 
 Reconstruct a geometry or feature using the function `f` on the `target` trait.
 
@@ -12,7 +21,11 @@ Reconstruct a geometry or feature using the function `f` on the `target` trait.
 
 The result is an functionally similar geometry with values depending on `f`
 
-# Flipped point the order in any feature or geometry, or iterables of either:
+$APPLY_KEYWORDS
+
+# Example
+
+Flipped point the order in any feature or geometry, or iterables of either:
 
 ```juia
 import GeoInterface as GI
--- a/src/transformations/extent.jl
+++ b/src/transformations/extent.jl
@@ -1,29 +1,17 @@
 """
     embed_extent(obj)
 
-Recursively wrap the object with a `GeoInterface.Wrappers` geometry,
+Recursively wrap the object with a GeoInterface.jl geometry,
 calculating and adding an `Extents.Extent` to all objects.
 
-This can improve performance when extents need to be checked multiple times.
-"""
-embed_extent(x; kw...) = apply(extent_applicator, GI.AbstractTrait, x; kw...)
+This can improve performance when extents need to be checked multiple times,
+such when needing to check if many points are in geometries, and using their extents
+as a quick filter for obviously exterior points.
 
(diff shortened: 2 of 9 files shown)


src/transformations/extent.jl
```diff
@@ -6,19 +6,24 @@ calculating and adding an `Extents.Extent` to all objects.
...
         embed_extent(g)
     end
     wrapper_type = GI.geointerface_geomtype(trait)
     extent = GI.extent(wrapper_type(children_with_extents))
-    return wrapper_type(children_with_extents, extent)
+    return wrapper_type(children_with_extents; extent, crs=GI.crs(geom))
 end
-extent_applicator(::PointTrait, point) = point
-extent_applicator(::PointTrait, point) = point
+function extent_applicator(trait::Union{GI.AbstractCurveTrait,GI.MultiPointTrait}, geom)
```
skygering, line 24: What happens with points? I guess it doesn't make a lot of sense for someone to call this on a point, but if they did, does it make sense to have something specifically defined for that?

src/transformations/extent.jl
```diff
@@ -6,19 +6,24 @@ calculating and adding an `Extents.Extent` to all objects.
 
 This can improve performance when extents need to be checked multiple times.
```
skygering, line 7: I think this could use a bit more comments. I wasn't familiar with this concept and it took a good few minutes to understand what was going on. Maybe a brief description of what extents are and how to access them after running.

src/transformations/extent.jl
rafaqz: Its just moved a few lines down and the dup deleted

src/transformations/extent.jl
```diff
@@ -6,19 +6,24 @@ calculating and adding an `Extents.Extent` to all objects.
...
 """
-embed_extent(x; kw...) = apply(AbstractTrait, x; kw...)
+embed_extent(x; kw...) = apply(extent_applicator, GI.AbstractTrait, x; kw...)
```
skygering, line 9: We also might want to specify here what keywords are possible. Cross-referencing with the with `primitives.jl`, I can see that it they are: `crs`, `calc_extent`, and `threaded`... however, down below we recalculate crs and we are obviously calculating extent so we might what to specify why we use apply.

test/transformations/extent.jl
```diff
@@ -0,0 +1,17 @@
+using Test
+ 
+import GeoInterface as GI
+import GeometryOps as GO
```
skygering, line 4: Would be good to test at least one curve or something that isn't nested since that is a separate function

skygering:
It seems like the `Extents` package isn't installed so the tests are failing. Also, we are getting a: `Could not use exact versions of packages in manifest` warning.

src/transformations/extent.jl
rafaqz: Ahh right, I'm assuming a lot of knowledge here. Definately more docs could help!

The ideas is you can use the extent as e.g. a fast `pointinpolygon` proxy. If its in the extent you need to actually check the polygon, if its not you know in a few ns.

src/transformations/extent.jl
rafaqz: Ahh probably this can all be written just using one line - `apply(identity, PointTrait, x; `calc_extent=true)` !!!

`calc_extent` already preopagates the extents efficiently.

And then the `threaded` and `crs` keywords make sense. We can just block the `calc_extent` keyword.  

lol

test/transformations/extent.jl
rafaqz: If we use `apply` for everything than it also needs less tests.

rafaqz:
Ok the code is so simple now this PR is more about adding keyword documentation to all the transformations.

I thought I may as well do that systematically if I added it to `embed_extent`.

src/primitives.jl
```diff
@@ -1,18 +1,31 @@
+const THREADED_KEYWORD = "- `threaded`: `true` or `false`. Whether to use multithreading. Defaults to `false`."
```
skygering, line 1: Oh this is such a good way of not having to retype the same documentation over and over. I am so stealing this for my other project 😆

skygering:
It is looking really good. I think it might be worth it in some PR, even if not this PR, to clean up the `apply` file and add some comments. It is not easy to skim and get a grasp on what is going on. But that can be a different PR if you think it is out of scope.

src/primitives.jl
rafaqz: Yeah, from Rasters.jl experience. Actually the main reason is so users see exactly the same things everywhere and it feels consistent

rafaqz:
Yeah I'll add that later

PR #30, opened 2023-12-26
better primitives docs and comments, and some tweaks

rafaqz:
@skygering let me know if this makes more sense to you.

Explaining recursive methods is kind of hard, and I write a lot of code like this so I probably don't see how or what is confusing.

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -105,9 +105,9 @@ function _apply(f, ::Type{Target}, ::Nothing, A::AbstractArray; threaded=false, 
     end
 end
 # There is no trait and this is not an AbstractArray.
-# Try to aclla _apply over. We can't use threading on an 
-# arbitrary iterable as we maybe can't index into it. So just `map`.
-# (TODO: maybe `collect` if `threaded=true` so we can thread?)
+# Try to call _apply over it. We can't use threading 
+# as we don't know if we can can index into it. So just `map`.
+# (TODO: maybe `collect` first if `threaded=true` so we can thread?)
 _apply(f, ::Type{Target}, ::Nothing, iterable; kw...) where Target =
     map(x -> _apply(f, Target, x; kw...), iterable)
 # Rewrap all FeatureCollectionTrait feature collections as GI.FeatureCollection
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -104,8 +104,10 @@ function _apply(f, ::Type{Target}, ::Nothing, A::AbstractArray; threaded=false, 
         _apply(f, Target, A[i]; threaded=false, kw...)
     end
 end
-# Try to _apply over unknown iterables. We can't use threading on an 
+# There is no trait and this is not an AbstractArray.
+# Try to aclla _apply over. We can't use threading on an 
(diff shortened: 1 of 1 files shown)


src/primitives.jl
```diff
@@ -12,14 +12,67 @@ $CALC_EXTENT_KEYWORD
 
 # This file mainly defines the [`apply`](@ref) function.
 
+#=
+## What is `apply`?
+
+`apply` applies some function to every geometry matching the `Target`
+GeoInterface trait, in some arbitrarily nested object made up of:
+- `AbstractArray`s 
+- Some arbitrary iterables may also work here
```
skygering, line 21: I am not really sure what this is referring too

skygering:
I did a spell check, but yeah other than my one comment above, I get what it does. Reads pretty well

src/primitives.jl
rafaqz: You mean the dot points? They are the types that will be accepted and recursed over by `apply`.

Maybe I can make that clearer.

rafaqz: Or you mean the iterables? it doesn't quite sit with the others but anything unknown (not trait and not an AbstractArray) we try to iterate over.

skygering: Okay yeah that makes sense.

skygering:
Happy for it to merge

src/primitives.jl
rafaqz: Ok I'll try to work that into the text

rafaqz, line 21:
```suggestion
- `AbstractArray`s (we also try to iterate other non-GeoInteface compatible object)
```

src/primitives.jl
```diff
@@ -12,14 +12,66 @@ $CALC_EXTENT_KEYWORD
...
+## What is `apply`?
+
+`apply` applies some function to every geometry matching the `Target`
+GeoInterface trait, in some arbitrarily nested object made up of:
+- `AbstractArray`s (we also try to iterate other non-GeoInteface compatible object)
+- `FeatureCollectionTrait` objects
+- `FeatureTrait` objects
+- `AbstractGeometryTrait` objects
+
+It recursively calls `apply` through these nested
```
rafaqz, line 25:
```suggestion
`apply` recursively calls itself through these nested
```

src/primitives.jl
```diff
@@ -12,14 +12,66 @@ $CALC_EXTENT_KEYWORD
...
+
+`apply` applies some function to every geometry matching the `Target`
+GeoInterface trait, in some arbitrarily nested object made up of:
+- `AbstractArray`s (we also try to iterate other non-GeoInteface compatible object)
+- `FeatureCollectionTrait` objects
+- `FeatureTrait` objects
+- `AbstractGeometryTrait` objects
+
+`apply` recursively calls itself through these nested
+layers until it reaches the `Target`, where it applies `f`, and stops.
```
rafaqz, line 26:
```suggestion
layers until it reaches objects with the `Target` GeoInterface trait. When found `apply` applies the function `f`, and stops.
```

Issue #32, opened 2023-12-27
Area LibGEOS Polygon problems


skygering:
I think the LibGEOS memory leak is causing havok far and wide. I was just doing some quick benchmarks on the new `area` code and just about had a heart attack.  I was using this polygon:
```
import LibGEOS as LG
import GeoInterface as GI
import GeometryOps as GO
using BenchmarkTools
pLG = LG.Polygon([
    [
        [0.0, 5.0], [2.0, 2.0], [5.0, 2.0], [2.0, -2.0], [5.0, -5.0],
        [0.0, -2.0], [-5.0, -5.0], [-2.0, -2.0], [-5.0, 2.0], [-2.0, 2.0],
        [0.0, 5.0],
    ],
])

pGI = GI.Polygon([
    [
        (0.0, 5.0), (2.0, 2.0), (5.0, 2.0), (2.0, -2.0), (5.0, -5.0),
        (0.0, -2.0), (-5.0, -5.0), (-2.0, -2.0), (-5.0, 2.0), (-2.0, 2.0),
        (0.0, 5.0),
    ],
])

GO.equals(pLG, pGI)  # true

@benchmark LG.area(pLG)
@benchmark GO.area(pLG)
@benchmark GO.area(pGI)
```
I got the following:
<img width="843" alt="Screenshot 2023-12-27 at 2 49 01 PM" src="https://github.com/asinghvi17/GeometryOps.jl/assets/60117338/9ad51476-861d-4404-b173-25c5cd8a06ca">

So it seems like the method itself isn't bad, but it does NOT play nice with LibGEOS.

rafaqz:
It could be type instability too

evetion:
This was runtime dispatch on GeoInterface.x and y. Solved by https://github.com/JuliaGeo/LibGEOS.jl/pull/191, which brings the benchmark to 4μs for me (still slow, but it's copying points from C, so expected).

rafaqz:
That's not time for copying points from C - it's mostly for allocating arrays in Julia, the yellow `Array`

evetion:
> That's not time for copying points from C - it's mostly for allocating arrays in Julia, the yellow `Array`

That picture was the before state. The afterward profile is in the other issue in LibGEOS.

rafaqz:
It still has those allocations tho ;)

rafaqz:
Ok right, I think there are still a bunch of allocations happening though when each point is loaded.

[...]

PR #33, opened 2023-12-27
Sg/area dist debug

skygering:
I realized that I missed a few geometry types that were coming up in my simulations.

I also realized that my code failed with empty geometries as it was trying to access the first point. This then led me to the discovery of an issue with the implementation of `isempty` at least with LibGEOS. So now I have to check `isempty` and find if there is a non-zero number of points to determine the type of the points within a geometry that returns an area of 0.

```
import LibGEOS as LG
import GeoInterface as GI

p = LG.readgeom("POLYGON EMPTY")
GI.isempty(p)  # false
GI.npoint(p) # 0
```

This seems like too much work just to return 0. I am wondering if it would be bad to just return an `Int` 0 for everything but non-empty polygons / multipolygon / collections. 

This isn't the best type-wise, but there is no guarantee that we can even determine the type. 

Thoughts?


src/methods/area.jl
```diff
@@ -71,23 +71,53 @@ computed slighly differently for different geometries:
...
+area(::GI.PointTrait, point) = GI.isempty(point) ?
+    0 : zero(typeof(GI.x(point)))
 
 signed_area(trait::GI.PointTrait, point) = area(trait, point)
 
+# MultiPoints
+function area(::GI.MultiPointTrait, multipoint)
+    GI.isempty(multipoint) && return 0
+    np = GI.npoint(multipoint)
+    np == 0 && return 0
```
rafaqz, line 83: Could we make area always return `Float64` for type stability?

Another option is to check the type once at the start, then push it through as the first argument to an `_area` method, and let users specify it too if they need to

src/methods/area.jl
```diff
@@ -71,23 +71,53 @@ computed slighly differently for different geometries:
...
+
+signed_area(trait::GI.MultiPointTrait, multipoint) = area(trait, multipoint)
+
 # Curves
-area(::CT, curve) where CT <: GI.AbstractCurveTrait =
-    zero(typeof(GI.x(GI.getpoint(curve, 1))))
+function area(::CT, curve) where CT <: GI.AbstractCurveTrait
+    GI.isempty(curve) && return 0
+    np = GI.npoint(curve)
+    np == 0 && return 0
```
rafaqz, line 93: Same here

rafaqz (commented):
I think we need to either pick a floating point output type, or detect it once very early and push that through. Then all those zeros will be type stable and we cont need to do the `zero(typeof(x(...`

skygering:
I wanted to preserve the flexibility so that if I run a simulation with Float32 or Float64 polygons, the area returns the correct type. But I can see why that is tricky, especially with the empty geometries for which we would have to default to something. Do you think it is better just for it to return `Float64` so that the user can plan?

rafaqz:
Maybe default to Float64 but let the user pass in the type?

skygering:
Okay, I updated the type stuff! They all have optional parameters now. I also just simplified some stuff.

Something I am noting is that I am doing `GI.getpoint(geom, i)` a lot in order to deal with "closed" curves. This is my way of making it so that it doesn't matter if a Geometry library repeats the last coordinates in linear rings / polygons, we still repeat the last point for our algorithms. I don't have an idea for a better way of doing this though. 

As discussed in #32, this can be problematic if a library does a poor job of implementing GeoInterface. I kinda feel like that is the library's fault though, not ours. For example, using `tuples` to convert an input geometry into a GeoInterface geometry feels like cheating, because we aren't really dealing with an arbitrary polygon, we're converting a given polygon to a specific type... 

However, if you think this is still slower than `GI.getpoint(geom)` and using the iterator regardless, then we can think up a new way of iterating. Maybe that can be a separate PR though.

rafaqz (approved):
Looks like it will be much faster :)

Lets not worry about `getgeom` for now. But I do worry that getting individual points from LibGeos or ArchGDAL will _never_ be fast.

PR #35, opened 2023-12-28
add transform method for coordinate transformations

### LINKED ISSUE #34: Support CoodinateTransformations.jl and Rotations.jl
This should be very little code, but it would be great to have a `transform` method that just uses `apply` to transform whole geometries with CoordinateTransformation.jl, Rotations.jl, and other packages that extend those or work together. 

It would be nice for users if that "just worked" with a very simple api.

rafaqz:
Closes #34 

This PR adds a simple `transform` function that converts points to `SVector` on the fly, transforms them using some transformation from CoordinateTransformations.jl or Rotations.jl, and rebuilds the object around the new points.

Like a few of the other transformation methods, this function is a pretty basic use of `apply`. But I think it may not seem obvious to people that you can do this so easily so it's worth having.

### DIFF
--- a/src/transformations/transform.jl
+++ b/src/transformations/transform.jl
@@ -42,11 +42,11 @@ ctor{2, Int64}[[2, 1], [4, 3], [6, 5], [2, 1]], nothing, nothing), GeoInterface.
 function transform(f, geom; kw...) 
     if _is3d(geom)
         return apply(PointTrait, geom; kw...) do p
-            f(StaticArrays.SVector{3}((GI.y(p), GI.x(p), GI.z(p))))
+            f(StaticArrays.SVector{3}((GI.x(p), GI.y(p), GI.z(p))))
         end
     else
         return apply(PointTrait, geom; kw...) do p
-            f(StaticArrays.SVector{2}((GI.y(p), GI.x(p))))
+            f(StaticArrays.SVector{2}((GI.x(p), GI.y(p))))
         end
     end
 end
--- a/src/transformations/transform.jl
+++ b/src/transformations/transform.jl
@@ -1,0 +1,52 @@
+
+"""
+    transform(f, obj)
+
+Apply a function `f` to all the points in `obj`.
+
+Points will be passed to `f` as an `SVector` to allow
+using CoordinateTransformations.jl and Rotations.jl 
+without hassle.
+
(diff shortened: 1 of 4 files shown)


rafaqz:
Any comments on this @skygering

src/transformations/transform.jl
```diff
@@ -0,0 +1,52 @@
...
+```julia
+julia> geom = GI.Polygon([GI.LinearRing([(1, 2), (3, 4), (5, 6), (1, 2)]), GI.LinearRing([(3, 4), (5, 6), (6, 7), (3, 4)])]);
+
+julia> f = CoordinateTransformations.Translation(3.5, 1.5)
+Translation(3.5, 1.5)
+
+julia> GO.transform(f, geom)
+GeoInterface.Wrappers.Polygon{false, false, Vector{GeoInterface.Wrappers.LinearRing{false, false, Vector{StaticArraysCore.SVector{2, Float64}}, Nothing, Nothing}}, Nothing, Nothing}(GeoInterface.Wrappers.Linea
+rRing{false, false, Vector{StaticArraysCore.SVector{2, Float64}}, Nothing, Nothing}[GeoInterface.Wrappers.LinearRing{false, false, Vector{StaticArraysCore.SVector{2, Float64}}, Nothing, Nothing}(StaticArraysCo
+re.SVector{2, Float64}[[5.5, 2.5], [7.5, 4.5], [9.5, 6.5], [5.5, 2.5]], nothing, nothing), GeoInterface.Wrappers.LinearRing{false, false, Vector{StaticArraysCore.SVector{2, Float64}}, Nothing, Nothing}(StaticA
```
skygering, line 25: The explanation here is great! I would not have known that I could do this. Is it efficient? I can use this in my code. 

However, the polygon in the example seems like it was translated by (4.5, 0.5) given the output. Unless I don't understand what the function is doing.

skygering: Yeah, the output is different than in the test, which seems to have the correct output for (3.5, 1.5).

skygering (approved):

skygering:
@rafaqz  Looks good other than my one comment. Sorry for the delay!

src/transformations/transform.jl
rafaqz: Oh yeah the example is from when there was a bug, nice catch!

PR #36, opened 2024-01-01
faster area

rafaqz:
With all the fixes in LibGEOS this gets us to ~20x slower territory for #32 

```julia
julia> @benchmark LG.area($pLG)
BenchmarkTools.Trial: 10000 samples with 990 evaluations.
 Range (min … max):  32.888 ns … 71.911 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     33.754 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   35.655 ns ±  4.169 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆▄█▇ ▄▁▁▄  ▂             ▁    ▂ ▁▄▂   ▃    ▁                ▁
  █████████████▇▇█▄▇█▅▅▆▇▅▇█▅▄▃▆█▆███▆▆██▇▅▆▄█▇▄▅▆▆▆▅▅▅▇▅▆▅▄▅ █
  32.9 ns      Histogram: log(frequency) by time      49.9 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark GO.area($pLG)
BenchmarkTools.Trial: 10000 samples with 143 evaluations.
 Range (min … max):  714.762 ns …  5.840 ms  ┊ GC (min … max): 0.00% … 5.31%
 Time  (median):       1.111 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):     2.401 μs ± 73.988 μs  ┊ GC (mean ± σ):  2.36% ± 0.08%

    ▁█ ▁  ▁      ▃▂                                             
  ▁▂██▅█▆▅█▅▄▄▅▆▆██▃▃▂▂▁▂▂▃▃▃▃▃▃▃▃▃▂▂▂▂▂▂▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁ ▂
  715 ns          Histogram: frequency by time         2.45 μs <

 Memory estimate: 80 bytes, allocs estimate: 3.

julia> @benchmark GO.area($pGI)
BenchmarkTools.Trial: 10000 samples with 996 evaluations.
 Range (min … max):  18.195 ns … 461.430 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     18.467 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   22.606 ns ±  12.344 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █   ▁ ▁▂   ▃ ▃ ▅ ▅  ▁ ▂  ▁                       ▁▁▁▁▂▂▂▁▁   ▂
  █▇█▁██████▆█▆█████▅▇█▆█▇▆███▄▅▅▆▆▆▅▆▆▆▇▇▆▆▇███▇█████████████ █
  18.2 ns       Histogram: log(frequency) by time      37.9 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

(and 2x faster than libgeos when working with native geoms!)

The idea with deleting all the handling of closed/open polygons is that the last point will contribute zero area anyway if its a repeat of the first one - so we can just ignore it.

### DIFF
--- a/src/methods/area.jl
+++ b/src/methods/area.jl
@@ -120,16 +120,16 @@ to be closed.
 =#
 function _signed_area(::Type{T}, geom) where T
     area = zero(T)
-    # Close curve, even if last point isn't explicitly repeated
     np = GI.npoint(geom)
     np == 0 && return area
+
     first = true
     local pfirst, p1
     # Integrate the area under the curve
     for p2 in GI.getpoint(geom)
         # Skip the first and do it later 
-        # This lets us work withing one iteration, which means on C call when 
-        # using points from external libraries
+        # This lets us work within one iteration over geom, 
+        # which means on C call when using points from external libraries.
         if first
             p1 = pfirst = p2
             first = false
@@ -137,7 +137,10 @@ function _signed_area(::Type{T}, geom) where T
         end
         # Accumulate the area into `area`
         area += GI.x(p1) * GI.y(p2) - GI.y(p1) * GI.x(p2)
+        p1 = p2
     end
+    # Complete the last edge.
+    # If the first and last where the same this will be zero
     p2 = pfirst
     area += GI.x(p1) * GI.y(p2) - GI.y(p1) * GI.x(p2)
     return T(area / 2)
--- a/src/methods/area.jl
+++ b/src/methods/area.jl
@@ -123,12 +123,22 @@ function _signed_area(::Type{T}, geom) where T
     # Close curve, even if last point isn't explicitly repeated
     np = GI.npoint(geom)
     np == 0 && return area
+    first = true
+    local pfirst, p1
     # Integrate the area under the curve
-    p1 = GI.getpoint(geom, np)
     for p2 in GI.getpoint(geom)
+        # Skip the first and do it later 
+        # This lets us work withing one iteration, which means on C call when 
+        # using points from external libraries
+        if first
+            p1 = pfirst = p2
+            first = false
+            continue
+        end
         # Accumulate the area into `area`
         area += GI.x(p1) * GI.y(p2) - GI.y(p1) * GI.x(p2)
-        p1 = p2
     end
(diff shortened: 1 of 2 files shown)


src/methods/area.jl
```diff
@@ -120,18 +120,28 @@ to be closed.
...
 function _signed_area(::Type{T}, geom) where T
     area = zero(T)
-    # Close curve, even if last point isn't explicitly repeated
     np = GI.npoint(geom)
     np == 0 && return area
-    first_last_equal = equals(GI.getpoint(geom, 1), GI.getpoint(geom, np))
-    np -= first_last_equal ? 1 : 0 
+
+    first = true
+    local pfirst, p1
```
skygering, line 127: What does this do? Just for my future knowledge?

src/methods/area.jl
```diff
@@ -120,18 +120,28 @@ to be closed.
...
     # Integrate the area under the curve
-    p1 = GI.getpoint(geom, np)
-    for i in 1:np
-        p2 = GI.getpoint(geom, i)
+    for p2 in GI.getpoint(geom)
+        # Skip the first and do it later 
+        # This lets us work within one iteration over geom, 
+        # which means on C call when using points from external libraries.
+        if first
```
skygering, line 133: Ah this is a pretty good way to do this.

skygering (approved):

skygering:
Looks great! Dang, the LibGEOS performance is still painful! But excited about the improvements with native polygons!

src/methods/area.jl
rafaqz: It makes the variables available in the scope outside thep loop

rafaqz:
I got it down to 3x faster than that with other changes...

PR #43, opened 2024-01-04
Sg/cross overlap

skygering:
Updates `crosses` and `overlaps` functions to use DE-9IM functions.

These don't work quite right at the moment. `Overlaps` needs to work even if there are intersections of dimension 0, as long as there are other intersections of dimension 1.


skygering:
Just wanted to get this out here in draft form so that I don't forget about it, even though I need to work on other things for the time being.

PR #47, opened 2024-01-18
Fix simplify bug

skygering:
Switch equals minus to minus equals


skygering:
Are we okay with just merging small bug fixes like this without review at this stage?

rafaqz:
For sure just merge unless feedback might help design/algorithm

PR #53, opened 2024-01-27
Ll/polygon intersects

LanaLubecke:
This branch has the most updated polygon clipping functions!


rafaqz:
Amazing. Is this ready for code review?

skygering:
Yep! Go for it! I have already reviewed it offline 😄  this is my intern @LanaLubecke !

src/methods/clipping/clipping_processor.jl
```diff
@@ -0,0 +1,281 @@
...
+                            n_new_per_poly += n_new_polys - 1
+                        end
+                    end
+                end
+            end
+        end
+        n_polys += n_new_per_poly
+    end
+    # Remove all polygon that were marked for removal
+    filter!(!isnothing, return_polys)::Vector{GI.Polygon}
```
rafaqz, line 279: This type assert is for an abstract type

src/methods/clipping/difference.jl
```diff
@@ -0,0 +1,83 @@
...
+
+## Example 
+
+```jldoctest
+import GeoInterface as GI, GeometryOps as GO
+
+poly1 = GI.Polygon([[[0.0, 0.0], [5.0, 5.0], [10.0, 0.0], [5.0, -5.0], [0.0, 0.0]]])
+poly2 = GI.Polygon([[[3.0, 0.0], [8.0, 5.0], [13.0, 0.0], [8.0, -5.0], [3.0, 0.0]]])
+diff_poly = GO.difference(poly1, poly2; target = GI.PolygonTrait)
+GI.coordinates.(diff_poly)
```
rafaqz, line 22: `GI.coordinates` is legacy we try not to encourage people to use it

rafaqz (changes requested):
Looks great, a few comments on style and performance.

Also wondering if this handles degeneracies? Like intersections exactly at vertices?

src/methods/clipping/difference.jl
skygering: What do you replace it with?

src/methods/clipping/clipping_processor.jl
```diff
@@ -0,0 +1,281 @@
+# # This file contains the shared helper functions for the polygon clipping functionalities.
+
+#= This is the struct that makes up a_list and b_list. Many values are only used if point is
+an intersection point (ipt). =#
+mutable struct PolyNode{T <: AbstractFloat}
```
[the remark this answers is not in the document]
skygering: It is mutable because you don't have all of the information for each node when they are made. Like you don't know where each node is in `b_list` when you are making `a_list`.  Would it be better to make them immutable and then just make a totally new node rather than updating the existing node?

src/methods/clipping/clipping_processor.jl
```diff
@@ -0,0 +1,281 @@
...
+    - Intersection: (x, y) -> x ? 1 : (-1)
+    - Difference: (x, y) -> (x ⊻ y) ? 1 : (-1)
+    - Union: (x, y) -> (x ⊻ y) ? 1 : (-1)
+
+A list of GeoInterface polygons is returned from this function. 
+=#
+function _trace_polynodes(a_list, b_list, a_idx_list, f_step)
+    n_a_pts, n_b_pts = length(a_list), length(b_list)
+    n_intr_pts = length(a_idx_list)
+    return_polys = Vector{GI.Polygon}(undef, 0)
```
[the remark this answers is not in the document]
skygering: Oh! How would you suggest getting around this? Making a vector with the first item as an object and then pushing? I don't understand the `GI.Polygon` type parameters...

src/methods/clipping/union.jl
```diff
@@ -0,0 +1,95 @@
+# #  Union Polygon Clipping
+export union
+
+"""
+    union(geom_a, geom_b, ::Type{T}; target::Type{Target})
```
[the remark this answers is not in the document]
skygering: Is this the docstring syntax for optional parameters then?

src/methods/clipping/clipping_processor.jl
skygering: Oops. I guess I was trying to make sure it knew it wasn't a type union with `Nothing`. If we filter out the `nothing` values, is that something we need to worry about?

src/methods/clipping/clipping_processor.jl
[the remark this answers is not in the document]
rafaqz: Probably worth benchmarking at least

src/methods/clipping/clipping_processor.jl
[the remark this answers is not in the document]
rafaqz: Sorry I made them pretty complicated. First two are hasz/is3d and hasm (for measures). Then it can wrap literally anything that is geointerface compatible or a nested vector of points

I guess construct one to get the type

src/methods/geom_relations/geom_geom_processors.jl
```diff
@@ -211,6 +211,7 @@ end
 the line (ls to le) and β is the traction of the curve (cs to ce). =#
 function _find_intersect_fracs(ls, le, cs, ce)
     _, fracs = _intersection_point(
+        Float64,
         (_tuple_point(ls), _tuple_point(le)),
```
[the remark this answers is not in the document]
LanaLubecke: What do you mean by everything else? Does this mean to not ignore the first output of `_intersection_point` and then use that in lines 221, 223, and 225?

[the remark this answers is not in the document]
skygering: I think @rafaqz means that we should save the `_tuple_point` outputs into a variable and then pass those tuples into all of the below functions rather than passing `ls`, `le`, `cs`, and `ce`. We could call the variables `t_ls`, `t_le`, ...

skygering:
It doesn't handle degeneracies. But neither does LibGEOS (https://github.com/JuliaGeo/LibGEOS.jl/issues/166) . There is an extension of the algorithm that we used that does and we are considering implementing the additional parts soon.

src/methods/geom_relations/geom_geom_processors.jl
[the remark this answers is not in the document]
LanaLubecke: I'm look at all the functions called in the while loop in line 128. Out of these, it seems only `segment_segment_orientation` (line 135) and `_find_intersect_fracs` (line 153) turn the inputs into tuples. So after creating `t_ls`, `t_le`, etc., these would be the only functions I pass the tuples into. Does that seem right?

skygering:
Just ran this on some pretty ugly polygons that @LanaLubecke designed that each have two overlapping holes. 

Screenshot 2024-01-28 at 5 41 48 PM

rafaqz:
Wild. We are so easily getting that order of magnitude faster.

src/methods/clipping/difference.jl
rafaqz: Honestly Idk 😭

In this case nothing is as simple. But you wouldnt use it in real code because the allocations are horrific (every point allocates a vector...)

I guess its fine for demonstrating the values

src/methods/clipping/clipping_processor.jl
rafaqz: Maybe try try spell out the full type?

But julia might work it out anyway. Profiling/Cthulhu should show it

skygering: I didn't see any slowdowns

src/methods/clipping/clipping_processor.jl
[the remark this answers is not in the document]
skygering: This doesn't quite work since we still need a type even if there aren't any polygons made. I hard-coded it for now and maybe we can add options for flexibility there later. But I would like to get this preliminarily merged before we dive too deep into that.

skygering:
I think we are good to go @rafaqz. Can you take one more look over it? Once it is merged, @LanaLubecke and I are going to have conversations about her next steps. That includes further profiling as well as implementing the degenerate cases.

.gitignore
```diff
@@ -1,4 +1,11 @@
 /Manifest.toml
 /docs/build/
 /docs/src/source/
-.vscode/
\ No newline at end of file
+.vscode/
+
```
rafaqz, line 5: Probably revert this file

rafaqz (approved):
Looks good. Degeneracy problems are fine for now if GEOS has them too 😂

We could also add a Martinez-Rueda alg at some stage to handke that
https://github.com/w8r/martinez

.gitignore
skygering: We aren't okay with having that in gitignore? I added it so that my .vscode files wouldn't get added by mistake.

rafaqz: No its all the lines below, sorry doesnt show in the comment

skygering: Oh yeah lol - that should go

PR #56, opened 2024-02-08
Sg/cut poly

skygering:
I needed this functionality in my code and figured that it might be useful to other people. While it isn't a feature in GEOS, it is a feature in MATLAB with the function [`cutpolygon`](https://www.mathworks.com/matlabcentral/fileexchange/24449-cutpolygon). It also has quite a few [similar posts on Stack Overflow](https://stackoverflow.com/questions/3623703/how-can-i-split-a-polygon-by-a-line). I was able to take advantage of the new clipping helper functions to write it. Hoping this isn't too specific. 

I also realized I made a type mistake in the clipping with the `Target` field so I fixed that as well.


skygering:
One thing I was wondering was if this should also have a Target field. To me, it seems intuitive that if you cut a polygon, you get a list of polygons. If you cut a multipolygon, you also get a list of polygons. The question is then how far would we want to extend something like this? If gets more confusing with cutting Linear Rings, for example as those would become LineStrings. Thoughts?

src/methods/clipping/cut.jl
```diff
@@ -0,0 +1,98 @@
+
```
asinghvi17, line 1:
```suggestion
# # Polygon cutting
# maybe a bit of context/explanation
# This is inspired by Matlab's [`cutpolygon`](https://www.mathworks.com/matlabcentral/fileexchange/24449-cutpolygon) function. 
# ## Example

```julia
import GeoInterface as GI, GeometryOps as GO
using CairoMakie
CairoMakie.activate!(pt_per_unit = 4); # hide

poly = GI.Polygon([[(0.0, 0.0), (10.0, 0.0), (10.0, 10.0), (0.0, 10.0), (0.0, 0.0)]])
line = GI.Line([(5.0, -5.0), (5.0, 15.0)])
cut_polys = GO.cut(poly, line)

f, a, p1 = poly(cut_polys; colors = [:blue, :orange])
lines!(a, line; color = :black)
f
```

# ## Implementation

skygering:
@asinghvi17 I updated the docs. Any other feedback?

src/methods/clipping/cut.jl
```diff
@@ -0,0 +1,132 @@
...
+cut_polys = GO.cut(poly, line)
+GI.coordinates.(cut_polys)
+
+# output
+2-element Vector{Vector{Vector{Vector{Float64}}}}:
+ [[[0.0, 0.0], [5.0, 0.0], [5.0, 10.0], [0.0, 10.0], [0.0, 0.0]]]
+ [[[5.0, 0.0], [10.0, 0.0], [10.0, 10.0], [5.0, 10.0], [5.0, 0.0]]]
+```
+"""
+cut(geom, line::GI.Line, ::Type{T} = Float64) where {T <: AbstractFloat} =
```
rafaqz, line 59: Is there a reason to limit to GI.Line here?

src/methods/clipping/difference.jl
```diff
@@ -28,7 +28,7 @@ GI.coordinates.(diff_poly)
 """
 function difference(
     geom_a, geom_b, ::Type{T} = Float64; target::Type{Target} = Nothing,
-) where {T <: AbstractFloat, Target <: GI.AbstractTrait}
+) where {T <: AbstractFloat, Target <: Union{Nothing, GI.AbstractTrait}}
```
rafaqz, line 31: Why `Nothing` here ?

src/methods/clipping/cut.jl
```diff
@@ -0,0 +1,132 @@
...
+    if n_intr_pts < 2 || isodd(n_intr_pts)
+        return [tuples(poly)]
+    end
+    # Cut polygon by line
+    cut_coords = _cut(T, ext_poly, poly_list, intr_list, n_intr_pts)
+    # Close coords and create polygons
+    for c in cut_coords
+        push!(c, c[1])
+    end
+    cut_polys = [GI.Polygon([c]) for c in cut_coords]
```
rafaqz, line 78: Could this be a `MultiPolygon`?

src/methods/clipping/cut.jl
```diff
@@ -0,0 +1,132 @@
...
+    # Add original polygon holes back in
+    _add_holes_to_polys!(T, cut_polys, GI.gethole(poly))
```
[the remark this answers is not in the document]
skygering: I don't think that is a problem. All this does is add back in "empty space" where there was originally space. If a hole is cut, then the edge of the exterior of the polygon is updated to be concave. Behind the scenes, this takes the difference of the polygon and the hole. I don't think we need to cut the hole before that since the polygon of interest won't extend beyond the cutting line anyways.

src/methods/clipping/difference.jl
skygering: Right now, if the target isn't specified then I have it set to `Nothing` as I am not sure what else it should be. I have a vision for that to change and have the input types determine the minimum viable union. So for example, if a target isn't provided and the inputs are polygons, then the target should be a union of points, lines, and polygons. But I haven't implemented that so the default is `Nothing` right now.

src/methods/clipping/cut.jl
rafaqz: I mean it could also accept other Lines, like GeometryBasics

src/methods/clipping/cut.jl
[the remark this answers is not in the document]
rafaqz: Ahh ok, was wondering how much happened tger. Perfect.

src/methods/clipping/cut.jl
skygering: Oh yeah, you're right. Whoops. I will fix that.

skygering:
I am going to merge this then now that I addressed those problems. Any disagreements?

PR #60, opened 2024-02-29
Coverage

skygering:
This calculates the area of a geometry within a bounding box defined by the minimum and maximum x and y values. So this should be similar to `coverage` within rasters, just for one polygon at a time. 

I haven't tested, but right now for multi-geometries and collections, I think it would be equivalent to Raster's `coverage` with the sum operation.

I am not sure if `area.jl` is the right place for this or if it should have its own file...


asinghvi17:
IMO coverage should probably be a separate file here?

skygering:
coverage_benchmarks

asinghvi17:
How many vertices does `p` have there?

rafaqz:
Could be good to accept an `Extents.Extent`

skygering:
> Could be good to accept an `Extents.Extent` for x/y min/max

Oh! That is a great idea. Then I don't need the point_in_cell functions! I will work swapping that!

skygering:
> How many vertices does `p` have there?

Not many. Like 10 or something. I deal with pretty small polygons (at least small numbers of points, if not area) most of the time so I don't have good "large" polygons to test on. If you have some and open a PR to put them in the test folder as an input that would be super cool.

skygering:
> > Could be good to accept an `Extents.Extent` for x/y min/max
> 
> Oh! That is a great idea. Then I don't need the point_in_cell functions! I will work swapping that!

@rafaqz In hindsight, I don't know if this is a good idea. On one hand, it would be a nicer input. On the other hand, we now have to make Extent objects, which is very, very cheap, but it still costs something. And if folks are dealing with cells, they would now need to make an Extent object for every cell. Thoughts? 

If we use: 
```julia
function f2(xmin, xmax, ymin, ymax, p)
       p_ext = GI.extent(p)
       cell_ext = Extents.Extent(X = (xmin, xmax), Y = (ymin, ymax))
       return Extents.intersects(p_ext, cell_ext)
end

f2(xmin, xmax, ymin, ymax, p) = xmin ≤ GI.x(p) ≤ xmax && ymin ≤ GI.y(p) ≤ ymax
```
then we get:
Screenshot 2024-03-04 at 12 52 27 PM

rafaqz:
It should cost nothing at all unless you jave mixed types somehow? They will promote. Otherwise its a bug.

I actually just meant we should let users pass in an extent if they want to

rafaqz:
Oh apctually by making the point an extent you are doubling the number of comparisons!!

skygering:
Okay, I moved the code and I also used `applyreduce` for multi-geometries. It is now type unstable due to that, but I am hoping that @asinghvi17's fix in his PR will fix that... 

But I feel good other than that. It can also take in an `extent` rather than the min/max values if desired.

rafaqz:
Whatever happend we will make `apply` and `applyreduce` type stable eventually

asinghvi17 (approved):
Looks good to me!  Once it's merged, I'll pull master into my branch and test it.

src/methods/clipping/coverage.jl
```diff
@@ -0,0 +1,301 @@
...
+This is the GeoInterface-compatible implementation. First, we implement a wrapper method
+that dispatches to the correct implementation based on the geometry trait. This is also used
+in the implementation, since it's a lot less work!
+
+Note that the coverage is zero for all points and curves, even if the curves are closed like
+with a linear ring.
+=#
+
+# Targets for applys functions
+const _COVERAGE_TARGETS = Union{GI.PolygonTrait,GI.AbstractCurveTrait,GI.MultiPointTrait,GI.PointTrait}
```
asinghvi17, line 40: Does coverage make sense for points or multi points?

src/methods/clipping/coverage.jl
```diff
@@ -0,0 +1,301 @@
...
+- The covverage of a geometry collection, multi-geometry, feature collection of
+    array/iterable is the sum of the coverages of all of the sub-geometries. 
+
+Result will be of type T, where T is an optional argument with a default value
+of Float64.
+"""
+function coverage(geom, xmin, xmax, ymin, ymax,::Type{T} = Float64; threaded=false) where T <: AbstractFloat
+    applyreduce(+, _COVERAGE_TARGETS, geom; threaded, init=zero(T)) do g
+        _coverage(T, GI.trait(g), g, T(xmin), T(xmax), T(ymin), T(ymax))
+    end
```
asinghvi17, line 64: This is the only thing I'm not sure about, since the function is inside the `applyreduce` call.  Still, it will operate at a pretty high level and hopefully that shouldn't cause too many issues.  

A potential solution might be to accept a tuple `(xmin, xmax, ymin, ymax)` in `_coverage`, and then just use `Base.Fix2` to create a function.

rafaqz: Giving the anonymous function a name just on the line above can help the compiler. Surprisingly. 

All the T types we are using everywhere worry me a but too, the compiler likes to forget what they are... sometimes it can help to wrap them in `Val` further out to force them through

src/methods/clipping/coverage.jl
skygering: I did what we did for `area` and it just returns zero for anything that isn't a polygon. I think it shouldn't error if you do supply in something other than a polygon, especially since you don't always know what types you're working with when some geometry functions return (for example LibGEOS functions).

src/methods/clipping/coverage.jl
skygering: I don't think I understand the problem. Can you elaborate? Is this something we need to fix pre-merge?

src/methods/clipping/coverage.jl
asinghvi17: Makes sense.

src/methods/clipping/coverage.jl
asinghvi17: This would be:
```suggestion
    coverage_for_subgeom(g) =  _coverage(T, GI.trait(g), g, T(xmin), T(xmax), T(ymin), T(ymax))
    applyreduce(coverage_for_subgeom, +, _COVERAGE_TARGETS, geom; threaded, init=zero(T))
```

which gets rid of the anonymous function created by the `do` block.

skygering: Should I change this in `area` as well then, since I copied this format from there?

asinghvi17: I can do that in my PR

asinghvi17:
I can merge this now and we'll get to the optimization in #64

PR #64, opened 2024-03-03
[WIP] Add geometry correction methods and benchmarks

asinghvi17:
Adds a framework for geometry correction methods, and a prototype for benchmarks against LibGEOS.

PR #69, opened 2024-03-05
Geometry corrections (unofficial dev branch)

asinghvi17:
These are the geometry correction and type stability changes forked out from #64.


src/methods/centroid.jl
```diff
@@ -169,7 +169,7 @@ end
 # It combines two (point, area) tuples into one, taking
 # the average of the centroid points weighted by the
 # area of the geom they are from.
-function _combine_centroid_and_area(((x1, y1), area1), ((x2, y2), area2))
+function _combine_centroid_and_area(((x1, y1), area1)#=::Tuple{Tuple{<: Real, <: Real}, <: Real}=#, ((x2, y2), area2)#=::Tuple{Tuple{<: Real, <: Real}, <: Real}=#) # ::Tuple{Tuple{<: Real, <: Real}, <: Real}
```
skygering, line 172: Is there a reason we can't put these in the comments above? Is this for documentation generation purposes? It would be easier to read above in my opinion. Also, can we make the comments above use multi-line comments? I have been doing that, again just for readability.

src/transformations/correction/valid_point.jl
```diff
@@ -0,0 +1,3 @@
+# # Valid Point
+
+# A point is valid if it has no `NaN` or `Inf` values.
```
skygering, line 3: Did we want code here? Or is this just setup for the next PR?

src/transformations/correction/closed_ring.jl
```diff
@@ -0,0 +1,66 @@
...
+# A closed ring is a ring that has the same start and end point. This is a requirement for a valid polygon (technically, for a valid LinearRing). 
+# This correction is used to ensure that the polygon is valid.
+
+# The reason this operates on the polygon level is that several packages are loose about whether they return LinearRings (which is correct) or LineStrings (which is incorrect) for the contents of a polygon.
+# Therefore, we decompose manually to ensure correctness.
+
+# ## Example
+#=
+
+Many polygon packages do not close their polygons, which makes them invalid according to strict specification.
```
skygering, line 14: I would add that this is also depended upon for many geometry algorithms, which is why this is usually the specification.

src/transformations/correction/valid_point.jl
asinghvi17: Ah this was just setup, but I can remove it for a cleaner PR.  This one was mainly for the API and structure - adding more should be simpler afterwards.

src/transformations/correction/closed_ring.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    return GI.Wrappers.Polygon([exterior, holes...])
+end
+
+function _close_linear_ring(ring)
+    if GI.getpoint(ring, 1) == GI.getpoint(ring, GI.npoint(ring))
+        # the ring is closed, all hail the ring
+        return ring
+    else
+        return GI.Wrappers.LinearRing([tuples.(GI.getpoint(ring))..., tuples(GI.getpoint(ring, 1))])
```
skygering, line 64: Isn't `GI.LinearRing` equivalent to `GI.Wrappers.LinearRing`?

src/methods/centroid.jl
asinghvi17: I think I was trying to remove a type instability and wrote this - thanks for catching it!  Will remove.

src/transformations/correction/closed_ring.jl
```diff
@@ -0,0 +1,66 @@
...
+"""
+struct ClosedRing <: GeometryCorrection end
+
+application_level(::ClosedRing) = GI.PolygonTrait
+
+function (::ClosedRing)(::GI.PolygonTrait, polygon)
+    exterior = _close_linear_ring(GI.getexterior(polygon))
+
+    holes = map(GI.gethole(polygon)) do hole
+        _close_linear_ring(hole) # TODO: make this more efficient, or use tuples!
```
skygering, line 53: Seems like you are using `tuples` below now... or do you mean a different use of tuples?

src/transformations/correction/closed_ring.jl
asinghvi17: Huh, I didn't know that!  I guess it never showed up in the REPL autocomplete so I forgot this was a thing.

src/transformations/correction/closed_ring.jl
asinghvi17: It still wraps the tuples in an array and does some splatting, which I do want to get rid of.

src/transformations/correction/geometry_correction.jl
```diff
@@ -0,0 +1,67 @@
...
+
+(gc::GeometryCorrection)(trait::GI.AbstractGeometryTrait, geometry) = error("Not implemented yet for $(gc) and $(trait).")
+
+function fix(geometry; corrections = GeometryCorrection[ClosedRing(),], kwargs...)
+    traits = application_level.(corrections)
+    final_geometry = geometry
+    for Trait in (GI.PointTrait, GI.MultiPointTrait, GI.LineStringTrait, GI.LinearRingTrait, GI.MultiLineStringTrait, GI.PolygonTrait, GI.MultiPolygonTrait)
+        available_corrections = findall(x -> x == Trait, traits)
+        isempty(available_corrections) && continue
+        println("Correcting for $(Trait)")
```
skygering, line 53: I mean like long term we don't want this feedback. But I am assuming this is for testing so.

skygering (commented):
Looks good overall! Excited to have this on its way. We should add some tests though.

src/transformations/correction/geometry_correction.jl
asinghvi17: Yep - I should probably add the `ValidPoint` check at least to test the interface, now that you mention it.

src/transformations/correction/closed_ring.jl
rafaqz: Yeah its in a module so you can import the whole lot into scopr if you want, but with `using` in GeoInterface so you can also just use `GI.Polygon`.

src/transformations/correction/closed_ring.jl
```diff
@@ -0,0 +1,68 @@
...
+
+    return GI.Wrappers.Polygon([exterior, holes...])
+end
+
+function _close_linear_ring(ring)
+    if GI.getpoint(ring, 1) == GI.getpoint(ring, GI.npoint(ring))
+        # the ring is closed, all hail the ring
+        return ring
+    else
+        return GI.Wrappers.LinearRing([tuples.(GI.getpoint(ring))..., tuples(GI.getpoint(ring, 1))])
```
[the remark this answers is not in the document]
asinghvi17: Good point - will change that!

test/transformations/correction/closed_ring.jl
```diff
@@ -0,0 +1,13 @@
+using Test
+
+import GeoInterface as GI
+import GeometryOps as GO
+
+open_rectangle = GI.Wrappers.Polygon([collect.([(0, 0), (10, 0), (10, 10), (0, 10)])])
+
+@testset "Closed Ring correction" begin
+    closed_rectangle = GO.ClosedRing()(open_rectangle)
```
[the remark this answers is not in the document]
rafaqz: Ah so this is one of many cirrections. I get it. Good for comppsing corrections. Maybe being able to pass a single cirrection as the first argument is more julian, like `rand` with rngs 

You might want them to be put in a Tuple rather than a vector too when theure grouped.

asinghvi17:
The type instability fix from here also exists in #74 so I'll make this one purely about geometry corrections

test/transformations/correction/closed_ring.jl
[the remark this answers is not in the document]
asinghvi17: You can already, since we only use `getindex` internally.

src/transformations/correction/valid_point.jl
```diff
@@ -0,0 +1,24 @@
...
+application_level(::ValidPoint) = Union{GI.LineStringTrait, GI.LinearRingTrait}
+
+function (::ValidPoint)(::Union{GI.LineStringTrait, GI.LinearRingTrait}, geom)
+    new_coords = NTuple{2, Float64}[]
+    sizehint!(new_coords, GI.npoint(geom))
+    for coord in GI.getpoint(geom)
+        x, y = GI.x(coord), GI.y(coord)
+        if isfinite(x) && isfinite(y)
+            push!(new_coords, (x, y))
+        end
```
asinghvi17, line 21: Is it better to iteratively delete elements from the array, rather than going through and reassembling?  Not sure here.  Will have to benchmark it...

rafaqz: Delete is pretty slow

asinghvi17: It's not guaranteed that the number of invalid points will be less than the number of valid points - consider a projected polygon outside of the current projection, for example...

rafaqz: Ok delete will be slow then. Maybe generating a bitindex is the fastest? I guess a benchmark is the only way to know

asinghvi17:
I guess this is the dev branch now :D.  Will have to rebase and merge.

src/transformations/correction/valid_point.jl
asinghvi17: I want to improve this but we can always benchmark later.

src/transformations/correction/geometry_correction.jl
asinghvi17:
```suggestion
        @debug "Correcting for $(Trait)"

**Comment by asinghvi17** [comment]
I think I'm pretty much done with what I want to accomplish here!  IMO this is mergeable as is - what do you think @rafaqz @skygering?

# Change log:
## New features
- Add a geometry correction framework with `fix` and one method, `ClosedRing()`, for it.  Others can be added in future.
- Add a `segmentize` method that corresponds to LibGEOS's density, that upsamples any geometry more complex than a line.
- Add a benchmark framework which runs quickly and well, and can plot on CI.
	- Implement proof of concept benchmarks for some functions (`simplify` and `segmentize`).
## Code changes
- Factor `reproject`'s implementation out to an extension on Proj
	- Provide a nice error hint if Proj isn't loaded and a user tries to use `reproject`.
- Since we're getting rid of Proj, also factor out the GeodesicSegments implementation for `segementize`.
- Various comment changes and documentation improvements.

File: src/transformations/segmentize.jl [hunk 10]
```diff
@@ -0,0 +1,219 @@
...
+- `max_distance::Real`: The maximum distance, **in meters**, between vertices in the geometry.
+- `equatorial_radius::Real=6378137`: The equatorial radius of the Earth, in meters.  Passed to `Proj.geod_geodesic`.
+- `flattening::Real=1/298.257223563`: The flattening of the Earth, which is the ratio of the difference between the equatorial and polar radii to the equatorial radius.  Passed to `Proj.geod_geodesic`.
+
+One can also omit the `equatorial_radius` and `flattening` keyword arguments, and pass a `geodesic` object directly to the eponymous keyword.
+
+This method uses the Proj/GeographicLib API for geodesic calculations.
+"""
+struct GeodesicSegments <: SegmentizeMethod 
+    geodesic# ::Proj.geod_geodesic
```
rafaqz, line 142: Is there a reason not to make this parametric?

asinghvi17: None at all!  I was not sure if the error hinting would work but it looks like it will.

PR #71, opened 2024-03-06
Line segmentization/densification methods

asinghvi17:
This PR adds line segmentization/interpolation methods for geometries.  I do want to extend this to n-D but that would make the algorithm slower, so might get to that later.

Currently, there are two segmentization types: linear and geodesic.  Linear is straight linear interpolation, geodesic is geodesic interpolation using the Proj geodesic API.

Still needs tests, which I will add in a bit.
download-8


src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+            t = i / n_segments
+            push!(new_coords, (x1 + t * dx, y1 + t * dy))
+        end
+    end
+    # End the line with the original coordinate,
+    # to avoid any multiplication errors.
+    push!(new_coords, (x2, y2))
+    return nothing
+end
+
```
rafaqz, line 158:
```suggestion
```

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+function segmentize(method::SegmentizeMethod, geom; threaded::Union{Bool, BoolsAsTypes} = _False())
+    @assert method.max_distance > 0 "`max_distance` should be positive and nonzero!  Found $(method.max_distance)."
+    segmentize_function = Base.Fix1(_segmentize, method)
+    return apply(segmentize_function, Union{GI.LinearRingTrait, GI.LineStringTrait}, geom; threaded)
+end
+
+_segmentize(method, geom) = _segmentize(method, geom, GI.trait(geom))
+
```
rafaqz, line 124:
```suggestion
```

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+
+
```
rafaqz, line 126:
```suggestion
```

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+- `method::SegmentizeMethod = LinearSegments()`: The method to use for segmentizing the geometry.  At the moment, only [`LinearSegments`](@ref) and [`GeodesicSegments`](@ref) are available.
+- `geom`: The geometry to segmentize.  Must be a `LineString`, `LinearRing`, or greater in complexity.
+- `max_distance::Real`: The maximum distance, **in the input space**, between vertices in the geometry.  Only used if you don't explicitly pass a `method`.
+
+Returns a geometry of similar type to the input geometry, but resampled.
+"""
+function segmentize(geom; max_distance, threaded::Union{Bool, BoolsAsTypes} = _False())
+    return segmentize(LinearSegments(; max_distance), geom; threaded)
+end
+
```
rafaqz, line 116:
```suggestion
```

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+One can also omit the `equatorial_radius` and `flattening` keyword arguments, and pass a `geodesic` object directly to the eponymous keyword.
+
+This method uses the Proj/GeographicLib API for geodesic calculations.
+"""
+struct GeodesicSegments <: SegmentizeMethod 
+    geodesic::Proj.geod_geodesic
+    max_distance::Float64
+end
+
```
rafaqz, line 93:
```suggestion
```

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+    @assert method.max_distance > 0 "`max_distance` should be positive and nonzero!  Found $(method.max_distance)."
+    segmentize_function = Base.Fix1(_segmentize, method)
+    return apply(segmentize_function, Union{GI.LinearRingTrait, GI.LineStringTrait}, geom; threaded)
+end
+
+_segmentize(method, geom) = _segmentize(method, geom, GI.trait(geom))
+
+
+
+# This is a
```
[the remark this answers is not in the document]
asinghvi17:
```suggestion
#= 
This is a method which performs the common functionality for both linear and geodesic algorithms, 
and calls out to the "kernel" function which we've defined per linesegment.
=#

File: src/transformations/segmentize.jl [hunk 7]
```diff
@@ -0,0 +1,226 @@
...
+
+_segmentize(method, geom) = _segmentize(method, geom, GI.trait(geom))
+
+
+
+# This is a 
+function _segmentize(method::Union{LinearSegments, GeodesicSegments}, geom, T::Union{GI.LineStringTrait, GI.LinearRingTrait})
+    first_coord = GI.getpoint(geom, 1)
+    x1, y1 = GI.x(first_coord), GI.y(first_coord)
+    new_coords = NTuple{2, Float64}[]
```
rafaqz, line 131: Why not assign the whole vector now? every point is pushed anyway

asinghvi17: The vector also holds the interpolated points, so pushing between the points would get bothersome I feel.

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+
+Returns a geometry of similar type to the input geometry, but resampled.
+"""
+function segmentize(geom; max_distance, threaded::Union{Bool, BoolsAsTypes} = _False())
+    return segmentize(LinearSegments(; max_distance), geom; threaded)
+end
+
+function segmentize(method::SegmentizeMethod, geom; threaded::Union{Bool, BoolsAsTypes} = _False())
+    @assert method.max_distance > 0 "`max_distance` should be positive and nonzero!  Found $(method.max_distance)."
+    segmentize_function = Base.Fix1(_segmentize, method)
```
skygering, line 117: What does this do? Does this prevent boxing or something?

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+This is useful for plotting geometries with a limited number of vertices, or for ensuring that a geometry is not too "coarse" for a given application.
+
+## Arguments
+- `method::SegmentizeMethod = LinearSegments()`: The method to use for segmentizing the geometry.  At the moment, only [`LinearSegments`](@ref) and [`GeodesicSegments`](@ref) are available.
+- `geom`: The geometry to segmentize.  Must be a `LineString`, `LinearRing`, or greater in complexity.
+- `max_distance::Real`: The maximum distance, **in the input space**, between vertices in the geometry.  Only used if you don't explicitly pass a `method`.
+
+Returns a geometry of similar type to the input geometry, but resampled.
+"""
+function segmentize(geom; max_distance, threaded::Union{Bool, BoolsAsTypes} = _False())
```
skygering, line 112: If someone passes in a `bool`, should we be making it into a `BoolAsTypes`?

rafaqz (approved):

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+    return apply(segmentize_function, Union{GI.LinearRingTrait, GI.LineStringTrait}, geom; threaded)
+end
+
+_segmentize(method, geom) = _segmentize(method, geom, GI.trait(geom))
+
+
+
+# This is a 
+function _segmentize(method::Union{LinearSegments, GeodesicSegments}, geom, T::Union{GI.LineStringTrait, GI.LinearRingTrait})
+    first_coord = GI.getpoint(geom, 1)
```
skygering, line 127: In other functions, we have also been letting the user pass in a type T that is either a Float32 or a Float64 so that the output types have that Float type rather than defaulting to Float64.

src/transformations/segmentize.jl
rafaqz: `apply` will do that

src/transformations/segmentize.jl
```diff
@@ -0,0 +1,226 @@
...
+        _fill_linear_kernel!(method, new_coords, x1, y1, x2, y2)
+        push!(new_coords, (x2, y2))
+        x1, y1 = x2, y2
+    end 
+    return rebuild(geom, new_coords)
+end
+
+function _fill_linear_kernel!(method::LinearSegments, new_coords::Vector, x1, y1, x2, y2)
+    dx, dy = x2 - x1, y2 - y1
+    distance = hypot(dx, dy) # this is a more stable way to compute the Euclidean distance
```
skygering, line 143: should we get rid of the Euclidean distance function then and use `hypot` everywhere?

skygering (commented):
Looks good in general. I just had a few questions about style things. 

Also, it would be good to start a test file that at least has your examples in there. Maybe just a test that there are more points and the `area` / `centroid` stays the same?

src/transformations/segmentize.jl
asinghvi17: `Fix1` wraps a function of 2 arguments with a constant first argument.  It's just a different way of saying
```julia
segmentize_function(x) = _segmentize(method, x)
```

src/transformations/segmentize.jl
asinghvi17: Ah, I'll have to take a look at that and duplicate the same logic here.

Out of curiosity are we also allowing BigFloats?

asinghvi17:
Yep - will do that!  I also have to introduce interpolation for z and m points which will make this a bit messier :D

src/transformations/segmentize.jl
skygering: I just made it a sub-type of abstract float so I guess that would work

src/methods/area.jl
```diff
@@ -97,6 +97,14 @@ _area(::Type{T}, ::GI.AbstractGeometryTrait, geom) where T = zero(T)
...
 _signed_area(::Type{T}, ::GI.AbstractGeometryTrait, geom) where T = zero(T)
 
+# LinearRings
+
+# For LinearRings, we default to the curve implementation.
+
+_area(::Type{T}, tr::GI.LinearRingTrait, geom) where T = abs(_signed_area(T, tr, geom))
+
+_signed_area(::Type{T}, ::GI.LinearRingTrait, geom) where T = _signed_area(T, geom)
+
```
asinghvi17, line 107:
```suggestion
# For LinearRings, we default to the curve implementation.

_area(::Type{T}, tr::GI.LinearRingTrait, geom) where T = 0 # could be abs(_signed_area(T, tr, geom))

_signed_area(::Type{T}, ::GI.LinearRingTrait, geom) where T = 0 # could be _signed_area(T, tr, geom)

PR #72, opened 2024-03-06
Remove ExactPredicates Dependance

skygering:
When I was profiling a case of `intersection` between two polygons with holes, I noticed that a LOT of time was spent on the calls to `ExactPredicates.meets`. It is generated code and for some reason didn't seem to be doing that efficiently.  That can be seen in the left and right (type stable) towers below:
Screenshot 2024-03-05 at 12 48 02 PM

Here was the performance with `ExactPredicates`:
Screenshot 2024-03-05 at 12 49 49 PM

I realized that I had code that was serving a similar purpose with `segment_segment_orientation` and that if I developed the `_intersection_points` code to cover all cases, then we had all of the needed functionality. After removing `ExactPredicates`, here is the performance of the same polygons:
Screenshot 2024-03-05 at 5 02 55 PM

You can see my issue on the type instability in the above profile in #73.


rafaqz:
ExactPredicates.jl shouldnt be that slow and def not type unstable. I was expecting a 2x slowown just fir the meets operarion, not the whole thing! (Unless thats 90% of the work...)

The thing about using it is it will get exact line intersections etc without floating point errors. That removes a bunch of edge case errors we could get otherwise, including returning invalid geometries.

But if it has serious problems we may need to involve the authors to fix them.

asinghvi17:
The conclusion from discussion on Slack is that it's probably OK to remove exact predicates for now, but we should add it back - either with an `exact` keyword at the top level, or a precision suggestion type which would separate methods in the same manner as other geometry libraries offer.  For example, CGAL offers `exact` v/s `fast` APIs.

src/methods/geom_relations/geom_geom_processors.jl
```diff
@@ -146,7 +137,7 @@ function _line_curve_process(
         for j in (closed_curve ? 1 : 2):nc
             c_end = _tuple_point(GI.getpoint(curve, j))
             # Check if line and curve segments meet
-            seg_val = _segment_segment_orientation((l_start, l_end), (c_start, c_end))
+            seg_val, intr1, _ = _intersection_point(Float64, (l_start, l_end), (c_start, c_end))
```
asinghvi17, line 140: For the future, maybe we could make this something like 
```julia
seg_val, intr1, _ = _intersection_point([ExactMath()/FastMath()], ...)
```
so that a math type indicator can be passed down smoothly.

rafaqz (approved):

asinghvi17:
Ah oops I already rebased and merged into #69.  Will patch these commits in as well!

asinghvi17:
Merging to main is fine - I should be able to ignore that code on the devbranch by merging main into it.  No worries!

Issue #66, opened 2024-03-04
Specifying `target` all the time in polygon methods is painful


asinghvi17:
Can we have it default to the greater of the two types given?

skygering:
I don't think so. Because if you give two `Polygons`, in LibGEOS it can return anything from a `Point` to a `LineString` to a `Polygon`. I think default behavior should probably mimic LibGEOS, which was my plan in the long term (but a list instead of a `GeometryCollection`). However, we could change that to do what you have suggested. It is more applicable for my use case. But then if people want LibGEOS behavior they have to specify that they want a `Union{Point, Line, LineString, Polygon} `which seems more annoying to me.

asinghvi17:
Ah, I see what you're saying.  I guess I wanted unspecified `target` to give the algorithm freedom to return anything, so the user doesn't have to specify at all.

skygering:
Ah yes. I agree with that. That is the long-term plan. So the default target would be` Union{GI.Point, GI.Line, GI.LineString, GI.Polygon} ` it is just that doesn't work right now, so I didn't want to give people a false understanding of what is possible with the current code.

PR #74, opened 2024-03-07
fix type stability of clipping methods

### LINKED ISSUE #73: Clipping Type Instabilities
I don't know if #69 takes care of this but, I am still seeing type instability, both from `tuples`, and also from the `target`.

Screenshot 2024-03-05 at 5 03 09 PM
Screenshot 2024-03-05 at 7 09 05 PM
Screenshot 2024-03-05 at 12 48 50 PM

To reproduce use:
```julia
import GeoInterface as GI
import GeometryOps as GO
using ProfileView
using Cthulhu

p25 = GI.Polygon([[(0.0, 0.0), (5.0, 0.0), (5.0, 8.0), (0.0, 8.0), (0.0, 0.0)],
    [(4.0, 0.5), (4.5, 0.5), (4.5, 3.5), (4.0, 3.5), (4.0, 0.5)],
    [(2.0, 4.0), (4.0, 4.0), (4.0, 6.0), (2.0, 6.0), (2.0, 4.0)]])
p26 = GI.Polygon([
    [(3.0, 1.0), (8.0, 1.0), (8.0, 7.0), (3.0, 7.0), (3.0, 5.0), (6.0, 5.0), (6.0, 3.0), (3.0, 3.0), (3.0, 1.0)],
    [(3.5, 5.5), (6.0, 5.5), (6.0, 6.5), (3.5, 6.5), (3.5, 5.5)],
    [(5.5, 1.5), (5.5, 2.5), (3.5, 2.5), (3.5, 1.5), (5.5, 1.5)]])

GO.intersection(p25, p26; target = GI.PolygonTrait)

f(n, p1, p2) = for _ in 1:n GO.intersection(p1, p2; target = GI.PolygonTrait) end
ProfileView.@profview f(10000, p25, p26)
```

The profile won't look just like the one above until #72 is merged, but the type unstable bar will still be there.

rafaqz:
Closes #73

Seems the problem is less complicated that we thought. `target` should be a constructed trait rather than a trait type mostly GeoInterfaces.jl does this too for the same reason.

2024-03-07-232703_963x982

I can't see any problems with `tuples`, the red block is just array allocation.


asinghvi17:
Should we do the same for apply?

rafaqz:
One reason not to do this in `apply` is you can't use `Union` or abstract types. I think that's why I wrote it like that. 

That might actually be a problem here too?

asinghvi17:
Hmm, that's a good point.  How about this:
```julia
struct Joint{T} 
end

Joint(args...) = Joint{Union{typeof.(args)...}()
```
and then you pass in a `Joint(GI.PointTrait(), GI.LineStringTrait())` to the type?  We can bikeshed the name, but that would be the general principle.

If better for general adoption we could even put this in GeoInterface.

rafaqz:
@skygering do you have any thoughts on changing this argument?

skygering:
Okay, so I originally had that. My only concern relates to @asinghvi17's comments in #66. In my opinion, this is more painful. However, I don't see it being that big of a problem given that we eventually will have default `MultiTraits` and people will only specify if they need one specific type out (or at least that is what I imagine being the biggest use case). We can also have some pre-defined MultiTraits that are the most common and useful. 

Basically, I don't love it, but if it improves performance, then I don't think it is _that_ bad. Does it improve runtime given the type instability doesn't propagate?

skygering:
> Hmm, that's a good point. How about this:
> 
> ```julia
> struct Joint{T} <: GeoInterface.AbstractTrait
> end
> 
> Joint(args...) = Joint{Union{typeof.(args)...}()
> ```
> 
> and then you pass in a `Joint(GI.PointTrait(), GI.LineStringTrait())` to the type? We can bikeshed the name, but that would be the general principle.
> 
> If better for general adoption we could even put this in GeoInterface.
> 
> Edit: an interesting name might be `MultiTrait` for this.

I like the name `MultiTrait` for this much more than `Joint`.

asinghvi17:
We could always convert a tuple input at the top level, and type info _should_ propagate through that, so I don't see it being too large of a burden?  It's also not that much different than typing `Union{...}` and quite a bit cleaner IMP

rafaqz:
I dont think `MultiTraits` will work in practice because we lose dispatch, that we keep with Unions

src/methods/clipping/difference.jl
```diff
@@ -27,7 +27,7 @@ GI.coordinates.(diff_poly)
 ```
 """
 function difference(
-    geom_a, geom_b, ::Type{T} = Float64; target::Type{Target} = Nothing,
+    geom_a, geom_b, ::Type{T} = Float64; target::Target = nothing,
 ) where {T <: AbstractFloat, Target <: Union{Nothing, GI.AbstractTrait}}
     return _difference(Target, T, GI.trait(geom_a), geom_a, GI.trait(geom_b), geom_b)
 end
```
skygering, line 33: Why doesn't the type on line 39 need to change to `::GI.PolygonTrait`, given it will be an object passed and not a type per say? (Sorry it won't let me comment on those lines for some reason)

asinghvi17: It's passing `Target` which is the type of `target`.

asinghvi17:
Summary from Slack:
* Acceptable forms of input to `target` are trait instances (`PointTrait()`), tuples of trait instances, union types, and multitraits.
* These are all converted to multitraits internally and propagated down.
* `MultiTrait`s are what we dispatch on even if only one trait is needed, by necessity.

is that the correct understanding of what we just discussed?

src/primitives.jl
```diff
@@ -2,6 +2,21 @@
...
 # This file mainly defines the [`apply`](@ref) function and its relatives.
 
+#=
+We pass `threading` and `calc_extent` as types, not simple boolean values.  
+
+This is to help compilation - with a type to hold on to, it's easier for 
+the compiler to separate threaded and non-threaded code paths.
+=#
+struct _True end
+struct _False end
```
asinghvi17, line 12:
```suggestion
abstract type BoolsAsTypes end
struct _True <: BoolsAsTypes end
struct _False <: BoolsAsTypes end
```
as in #69

rafaqz:
Ok, this PR is now implementing `TraitTarget` wrapper everywhere applicable. Its basically `Val` with a name. Has the same effect as switching to constructed traits.

src/primitives.jl
```diff
@@ -94,57 +110,48 @@ flipped_geom = GO.apply(GI.PointTrait, geom) do p
...
-We pass `threading` and `calc_extent` as types, not simple boolean values.  
-
-This is to help compilation - with a type to hold on to, it's easier for 
-the compiler to separate threaded and non-threaded code paths.
-=#
-struct _True end
-struct _False end
-
 @inline _booltype(x::Bool) = x ? _True() : _False()
 @inline _booltype(x::Union{_True,_False}) = x
```
asinghvi17, line 121: Can we move these definitions to where the booltypes are now defined, and change the last type from `Union` to `BoolsAsTypes`?

asinghvi17:
I haven't tried this out yet, but did you notice a performance speedup?

rafaqz:
The same one as swapping to constructed traits, shown in the profile.

`apply` is still not 100% type stable in all cases but I assume there is something else going on there.

src/primitives.jl
```diff
@@ -2,6 +2,22 @@
...
+abstract type BoolsAsTypes end
+struct _True <: BoolsAsTypes end
+struct _False <: BoolsAsTypes end
+
+# This struct holds a trait parameter or a union of trait parameters.
+struct TraitTarget{T} end
+TraitTarget(::Type{T}) where T = TraitTarget{T}()
+TraitTarget(::T) where T<:GI.AbstractTrait = TraitTarget{T}()
+TraitTarget(::TraitTarget{T}) where T = TraitTarget{T}()
+TraitTarget(::Type{<:TraitTarget{T}}) where T = TraitTarget{T}()
```
asinghvi17, line 20:
```suggestion
TraitTarget(::Type{<:TraitTarget{T}}) where T = TraitTarget{T}()
TraitTarget(traits::GI.AbstractTrait...) where T = TraitTarget{Union{typeof.(traits)...}}()

rafaqz: Does that compile away? You may want `map` instead of the broadcast.

PR #94, opened 2024-04-01
Implement `to_points` as `collect(flatten(...))`

asinghvi17:


src/utils.jl
```diff
@@ -95,31 +95,14 @@ function to_extent(edges::Vector{Edge})
     Extents.Extent(X=x, Y=y)
 end
 
-function to_points(x, ::Type{T} = Float64) where T
-    points = Vector{TuplePoint{T}}(undef, _npoint(x))
-    _to_points!(points, x, 1)
-    return points
-end
+_to_point3(T, p) = (T(GI.x(p)), T(GI.y(p)), T(GI.z(p)))
```
rafaqz, line 98: I think we should make these more official...

like `xy` and `xyz` without underscores, so we can use them everywhere.

src/utils.jl
```diff
@@ -95,31 +95,14 @@ function to_extent(edges::Vector{Edge})
...
-    for p in GI.getpoint(geom)
-        n += 1
-        points[n] = _tuple_point(p)
-    end
-    return n
+"""
+    to_points(geom::AbstractGeometry, ::Type{T} = Float64) where {T <: Number}
+"""
+function to_points(geom, ::Type{T} = Float64) where {T <: Number}
+    return collect(flatten(Base.Fix1(_is3d(geom) ? _to_point3 : _to_point2, T), GI.PointTrait, geom))
```
rafaqz, line 105: You probably want the conditional outside the generator... does this compile away to be only called once or is it called every time? 

If its outside we know it does. Its worth the duplication of the line anyway.

asinghvi17: I think that by definition it's called only once - the if statement is simply deciding which function to pass in, no?  It's not an if on the point but on the geometry.

rafaqz: No :)

Its inside a Fix2 so its called for every single point that `flatten` calls `f(geom)` on.

For ArchGDAL that will be painfully slow.

rafaqz, line 105: Deleted the others as I couldn't do a suggestion:
```suggestion
    if is3d(geom)
        return collect(flatten(p -> _to_point3(T, p)), GI.PointTrait, geom))
    else
        return collect(flatten(p -> _to_point2(T, p), GI.PointTrait, geom))
    end
```

This way `T` is type stable and and its more readable. Puttting a type in a Fix2 or as a closure argument doesn't work, its just `DataType`.

asinghvi17: Makes sense.  Thanks for the suggestion!

I didn't realize that the Fix2 doesn't propagate - that's a whoops.  It seemed to improve runtime earlier but that may have been a purely compile-time improvement.

rafaqz:
Would be good to see a performance comparison for this PR before mergin. It could be slow..

skygering:
This is much cleaner! I agree that a performance comparison would be nice!  

If it is faster, `_to_edges` is similarly long-winded for a very simple result. I wonder if we could do something similar.

asinghvi17:
Will check out the performance in a bit!

PR #95, opened 2024-04-02
convert to string rather than CRS in `reproject`

rafaqz:
Turns out it works just  converting everything to `String` rather than `Proj.CRS`.

This should be fixed in proj, but this seems fine for now.

### DIFF
--- a/ext/GeometryOpsProjExt/reproject.jl
+++ b/ext/GeometryOpsProjExt/reproject.jl
@@ -25,7 +25,7 @@ end
 function reproject(geom, source_crs, target_crs;
     time=Inf,
     always_xy=true,
-    transform=Proj.Transformation(Proj.CRS(source_crs), Proj.CRS(target_crs); always_xy),
+    transform=Proj.Transformation(convert(String, source_crs), convert(String, target_crs); always_xy),
     kw...
 )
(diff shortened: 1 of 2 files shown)


ext/GeometryOpsProjExt/reproject.jl
```diff
@@ -25,9 +25,16 @@ end
 function reproject(geom, source_crs, target_crs;
     time=Inf,
     always_xy=true,
-    transform=Proj.Transformation(convert(String, source_crs), convert(String, target_crs); always_xy),
+    transform=nothing,
     kw...
 )
+    transform = if isnothing(transform) 
+        s = source_crs isa Proj.CRS ? source_crs : convert(String, source_crs)
```
asinghvi17, line 32: Should this be something like `to_proj_consumable(source_crs)` instead, which can then have multiple dispatches?

asinghvi17: We can also add that later, since I want to figure out a good way to incorporate other transformation packages into this as well.

rafaqz: Yeah I thought about that too, but since its in a Proj extension right no thats not so easy

PR #99, opened 2024-04-06
[WIP] More geometry corrections

asinghvi17:
Add a couple of Julia corrections, and the long-awaited `GEOSCorrection` which does everything and its hat.

Still extremely WIP and not working in current form!


asinghvi17:
Stale but code can be copy pasted later

Issue #76, opened 2024-03-19
Implement what we can, wrap what we can't


asinghvi17:
We should add a method `GEOS()` to nearly every method here, which literally just calls `LibGEOS.(...)` so that we can expose a correct and working example of the feature.  This would be huge for adoption, since the GO interface is a lot more Julian and easy to use (plus accepts more types) than LG!

asinghvi17:
DelaunayTriangulation.jl already has a Graham scan algorithm for complex hull which we could hook into: https://github.com/JuliaGeometry/DelaunayTriangulation.jl/blob/ab58263640fc02e84053e5cc5e33a0fdc21d4fc7/src/data_structures/convex_hull.jl#L36-L43

asinghvi17:
This is probably understood pretty well by now, closing

PR #100, opened 2024-04-07
Implement a `GEOS` algorithm in an extension

asinghvi17:
This PR implements a parametrized `GEOS` algorithm within GeometryOps, and factors out the integration work to an extension `GeometryOpsLibGEOSExt`. (cf #76)

The extension is currently implemented for polygon set ops, DE-9IM functions, and buffering.

With this, you basically pass `GEOS(; params...)` as the first argument to whichever functions have been overridden.  Ideally, this should be at least most of the available functions.

```julia
GO.buffer(GEOS(), p1, 1)
```

Needs docs and tests.


ext/GeometryOpsLibGEOSExt/buffer.jl
```diff
@@ -0,0 +1,11 @@
+function GO.buffer(::GEOS, geom, distance)
+    return LG.buffer(GI.convert(LG, geom), distance)
+end
+
+function GO.buffer(::GEOS{(:cap_style, :join_style)}, geom, distance)
```
rafaqz, line 5: What if users only pass `cap_style`?

asinghvi17: Yet to be implemented :D - one of the things I haven't figured out yet is dispatch on subsets. Maybe we don't dispatch on keywords and test them internally instead?

ext/GeometryOpsLibGEOSExt/segmentize.jl
```diff
@@ -0,0 +1,26 @@
...
+    end
+    return LG.geomFromGEOS(result, context)
+end
+
+_segmentize_geos(geom, max_distance) = _segmentize_geos(GI.convert(LG, geom), max_distance)
+
+function GO.segmentize(alg::GEOS{(:max_distance,)}, geom; threaded::Union{Bool, GO.BoolsAsTypes} = _False())
+    return GO.apply(
+        Base.Fix2(_segmentize_geos, alg.params.max_distance), 
+        GO.TraitTarget(GI.MultiPolygonTrait(), GI.PolygonTrait(), GI.MultiLineStringTrait(), GI.LineStringTrait(), GI.LinearRingTrait(), GI.MultiPointTrait(), GI.PointTrait()),
```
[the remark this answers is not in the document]
asinghvi17:
```suggestion
        GO.TraitTarget(GI.GeometryCollectionTrait(), GI.MultiPolygonTrait(), GI.PolygonTrait(), GI.MultiLineStringTrait(), GI.LineStringTrait(), GI.LinearRingTrait(), GI.MultiPointTrait(), GI.PointTrait()),
```
It even accepts geometry collections, I just tested.

ext/GeometryOpsLibGEOSExt/buffer.jl
rafaqz: Yeah dispatch on keywords might not scale well, probably a conditional is easier.

asinghvi17:
I will have to work on this tomorrow but a basic framework of what has to happen:
- GEOS has to not be typed at all
- `get_defaults(geos, kwargs_input; defaults...)` or something that returns a namedtuple of defaults from the given kwargs.
- Potentially do misspelling checks, but that's not the goal here.

Sidenote: we need topological simplify as well from GEOS, that can be its own object.

rafaqz:
> GEOS has to not be typed at all

Can you explain this a bit

asinghvi17:
The parameter in the type can't exist, or more accurately it shouldn't be used for dispatch (which removes its purpose altogether)

rafaqz:
Well dispatch not the only purpose for type parameters... My concern would be where we are calling LibGEOS deep in `apply` and type instability in the repeated calls will matter quite a bit.

asinghvi17:
I guess that brings up another topic - should we eagerly convert back to GI.* or leave geometries as LibGEOS geoms?

rafaqz:
Its almost an arbitrary descision, we may need real world use to know either way. Eager conversion is clean and will make subsequent native julia ops faster, but always makes the GEOS op slower. And e.g. if people do repeated GEOS operations not converting will be fastest overall.

asinghvi17:
It does make things like segmentize type unstable unless you explicitly pass a method which is a bit strange...

asinghvi17:
@rafaqz this is what I was thinking of - in any case kwargs will not change dispatch behaviour I think, so the purpose they serve is limited.

rafaqz:
Ok if they dont affect types its less an issue. I guess just put LibGEOS calls in explicit conditionals to avoid dyamic dispatch slowness.

rafaqz:
Some tests would be nice ;)

Also just to see what everything actually does.

asinghvi17:
Seems fair.  I'm not going to test too thoroughly here since LibGEOS should be taking care of that, but can definitely test type combos + return types and keyword defaults/enforcement.

rafaqz:
At least make sure everything actually runs ;)

asinghvi17:
The PR should be ready for review now!

ext/GeometryOpsLibGEOSExt/buffer.jl
```diff
@@ -0,0 +1,29 @@
...
+
+to_cap_style(style::Symbol) = _GEOS_CAPSTYLE_LOOKUP[style]
+to_cap_style(style::LG.GEOSBufCapStyles) = style
+to_cap_style(num::Integer) = num
+
+to_join_style(style::Symbol) = _GEOS_JOINSTYLE_LOOKUP[style]
+to_join_style(style::LG.GEOSBufJoinStyles) = style
+to_join_style(num::Integer) = num
+
+function GO.buffer(alg::GEOS, geom, distance)
```
rafaqz, line 21: Shouldn't `buffer` use `apply` ? This wont work for `FeatureCollection`/`Feature`/iterables/tables

test/runtests.jl
```diff
@@ -1,5 +1,5 @@
 using GeometryOps
-using Test
+using Test, SafeTestsets
```
rafaqz, line 2: nice

ext/GeometryOpsLibGEOSExt/simple_overrides.jl
```diff
@@ -0,0 +1,60 @@
...
+
+This file contains simple overrides for GEOS, essentially only those
+functions which have direct counterparts in LibGEOS and only
+require conversion before calling.
+
+=#
+# ## Polygon set operations
+# ### Difference
+function GO.difference(::GEOS, geom_a, geom_b; target=nothing)
+    return LG.difference(GI.convert(LG, geom_a), GI.convert(LG, geom_b))
```
rafaqz, line 13: LG has these converts build in if you want to simplify this code a bit

src/transformations/segmentize.jl
```diff
@@ -169,12 +169,12 @@ This is useful for plotting geometries with a limited number of vertices, or for
 Returns a geometry of similar type to the input geometry, but resampled.
 """
 function segmentize(geom; max_distance, threaded::Union{Bool, BoolsAsTypes} = _False())
```
rafaqz, line 171:
```suggestion
@inline function segmentize(geom; max_distance, threaded::Union{Bool, BoolsAsTypes} = _False())
```

We should make sure to inline everything that has this bool/BoosAsTypes conversion so constprop works

ext/GeometryOpsLibGEOSExt/buffer.jl
```diff
@@ -0,0 +1,29 @@
...
+    :square => LG.GEOSBUF_CAP_SQUARE,
+)
+
+const _GEOS_JOINSTYLE_LOOKUP = Dict{Symbol, LG.GEOSBufJoinStyles}(
+    :round => LG.GEOSBUF_JOIN_ROUND,
+    :mitre => LG.GEOSBUF_JOIN_MITRE,
+    :bevel => LG.GEOSBUF_JOIN_BEVEL,
+)
+
+to_cap_style(style::Symbol) = _GEOS_CAPSTYLE_LOOKUP[style]
```
rafaqz, line 13:
```suggestion
_to_cap_style(style::Symbol) = _GEOS_CAPSTYLE_LOOKUP[style]
```

Nice to mark these little helpers as internals

src/transformations/segmentize.jl
asinghvi17: Will add a devdocs section in #108 about this as well then

ext/GeometryOpsLibGEOSExt/buffer.jl
asinghvi17: Good point...will adjust that.  At some point I should convert these to an eval loop, but that's probably out of scope for this PR

ext/GeometryOpsLibGEOSExt/simple_overrides.jl
asinghvi17: I've had issues before but will test if this works.  Possibly that was on simplify, which isn't implemented in LibGEOS

ext/GeometryOpsLibGEOSExt/buffer.jl
asinghvi17: It's not accessible in any case unless you use something like `Base.get_extension(GeometryOps, :GeometryOpsLibGEOSExt)` - so is that necessary in this case?

asinghvi17:
Oh geez...test failure is not present on my machine.  I guess allowing `Fix1`s to take arbitrary arguments was a 1.10 thing.

Will modify the tests accordingly.

ext/GeometryOpsLibGEOSExt/buffer.jl
rafaqz: Its more about signalling to the reader what the scope is. Its hard to know if there was an import somewhere just looking at these lines.

ext/GeometryOpsLibGEOSExt/simple_overrides.jl
rafaqz: Yeah, maybe I missed a few methods. Might be good to list those and we can fix them for everyone else too.

skygering: Yeah, I have hit a few of those. I believe `makeValid` and `isValid` only work with `LG` polygons as well.

src/transformations/correction/geos.jl
```diff
@@ -0,0 +1,22 @@
...
+    GEOSCorrection() <: GeometryCorrection
+
+This correction runs `LibGEOS.makeValid` on the highest level of each geometry available.
+
+See also [`GeometryCorrection`](@ref).
+"""
+struct GEOSCorrection <: GeometryCorrection end
+
+application_level(::GEOSCorrection) = TraitTarget(GI.MultiPolygonTrait(), GI.PolygonTrait(), GI.LinearRingTrait(), GI.MultiLineStringTrait(), GI.LineStringTrait(), GI.MultiPointTrait(), GI.PointTrait())
+
```
skygering, line 22: should we add in the actual `makeValid` call? Or is that in a different PR?

skygering: I guess I also don't really understand why this wouldn't be in the `LibGEOS` extensions folder since it is going to call `LibGEOS` code according to the comments and I would hope that one day we would have our own code for this.

skygering (approved):

src/transformations/correction/geos.jl
rafaqz: We cant add new methods or types in the extensions folder, they wont be visible to users. I think we need this file just for the `GEOScorrection` struct.

asinghvi17: This is a skeleton - I was going to add the actual correction in the extension, but the type has to be (a) <: AbstractGeometryCorrection and (b) accessible to users, which is not possible to do if it's implemented in the extension. 

The issue I ran into is that LG can return a multipolygon given polygon input, so I might have to reimplement apply for this with some hooks in the reconstructor...

skygering: Ah so the other ones work because they are also names of functions we have. Which is why we need the [not_implemented_yet.jl](https://github.com/JuliaGeo/GeometryOps.jl/pull/100/files#diff-14f503d5c8d2ca266096779fcd5ad1d48dd2667e71a52e0bce8cc12f9e85dcd6) file? Is that an accurate understanding?

asinghvi17: Yep, that's why `buffer` works for example

skygering: Also I just want to make sure that this one does return `GI.Polygon`. I feel like it would be confusing since it isn't one that we are passing `GEOS()` into, which would at least clearly mark that it would return an `LG` polygon.

skygering:
> Its almost an arbitrary descision, we may need real world use to know either way. Eager conversion is clean and will make subsequent native julia ops faster, but always makes the GEOS op slower. And e.g. if people do repeated GEOS operations not converting will be fastest overall.
> 
> Like your issue over at Rasters is caused by an unused eager conversion step in GDAL: [rafaqz/Rasters.jl#634](https://github.com/rafaqz/Rasters.jl/issues/634)
> 
> If GDAL could just pass points through to Proj unchanged there would be no problem.
> 
> Having GeoInterface we can avoid that. But I'm not sure its worth it on balance. Maybe a keyword can control it eventually, and users can choose not to convert back when they need the performance for something.

I like the idea of a keyword eventually. It really does depend on the use case and it is nice to have control over your types.

asinghvi17:
Maybe we could change the types at some point to have a `GEOSDirect` and `GEOSConvert` algorithm, where the results from LG functions go into `postprocess_results(::GEOSDirect, ...) = identity` and `postprocess_results(::GEOSConvert, ...) = GO.tuples(...)` or something.

asinghvi17:
The geometry correction seems a bit difficult to do correctly, so I'm considering removing it from this PR and fixing this up to merge as is.

asinghvi17:
I've had to add a hack in the test file for LG geometrycollection conversion to work, but all tests pass locally, and are using the GEOS backend for GeometryOps rather than LibGEOS directly.

IMO this should be ready for merge if tests pass.

src/transformations/correction/geos.jl
asinghvi17: GEOSCorrection has been removed for now to not complicate this PR.

ext/GeometryOpsLibGEOSExt/simplify.jl
```diff
@@ -0,0 +1,8 @@
+# Address potential ambiguities
+GO._simplify(::GI.PointTrait, ::GO.TopologyPreserve, geom; kw...) = geom
+GO._simplify(::GI.MultiPointTrait, ::GO.TopologyPreserve, geom; kw...) = geom
+
+function GO._simplify(::GI.AbstractGeometryTrait, alg::GO.TopologyPreserve, geom)
```
skygering, line 5: Why don't we have to pass in `:GEOS` here like we do in `simple_overrides.jl`? I get that we don't have a native way of doing that yet, but I worry that people who want to use the `GO.TopologyPreserve` algorithm won't even realize that is it calling `LibGEOS`. Perhapes this needs to be emphasized in comments here and (stronger) in comments in `simplify.jl`.

skygering: Should this have errors like buffer about LibGEOS needing to be loaded?

asinghvi17: GEOS has a regular `simplify` which is the equivalent of DouglasPeucker here, this is a different algorithm.  I can definitely add this to the docstring for the algorithm struct, though.

rafaqz: Yeah this is a little odd to add, especially when the other simplify methods all have keywords that `TopologyPreserve` doesn't have.

It would def need a "Load LibGEOS" warning.

test/runtests.jl
```diff
@@ -14,33 +14,60 @@ const AG = ArchGDAL
 const LG = LibGEOS
 const GO = GeometryOps
 
+# Hack LG to have GeometryCollection converts
+@eval LG begin
```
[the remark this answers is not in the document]
asinghvi17: LG does not have GI conversion to GeometryCollection yet, I have a PR up there to add it and will set compat here before a release.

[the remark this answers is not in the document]
asinghvi17: https://github.com/JuliaGeo/LibGEOS.jl/pull/207

skygering (commented):
Looks good overall! Just a few clarifying questions to address.

ext/GeometryOpsLibGEOSExt/simplify.jl
rafaqz: I think its best not to add this and let us to add our own TopologyPreseve later that doesn't mean GEOS topology preserve.

test/runtests.jl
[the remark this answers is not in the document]
rafaqz: GeometryCollection is missing from LG, ut probably PR over there is better? We need to do this for the testing PR anyway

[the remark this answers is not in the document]
rafaqz: I'll do it now lol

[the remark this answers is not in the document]
rafaqz: Ohh haha I already review that and its merged :smile:

ext/GeometryOpsLibGEOSExt/simplify.jl
rafaqz: Oh make it so you need to use `GEOS()` now as well as `TopologyPreserve`

rafaqz (approved):
LGTM

PR #102, opened 2024-04-10
Multipolygon Clipping

skygering:
Adds multipolygon clipping and a fix for invalid multipolygons due to overlaps in sub-polygons.

### DIFF
--- a/src/methods/clipping/difference.jl
+++ b/src/methods/clipping/difference.jl
@@ -100,6 +100,8 @@ tracing a_list, else step backwards, where x is the entry/exit status and y is a
 that is true if we are on a_list and false if we are on b_list. =#
 _diff_step(x, y) = (x ⊻ y) ? 1 : (-1)
 
+#= Polygon with multipolygon difference - note that all intersection regions between
+`poly_a` and any of the sub-polygons of `multipoly_b` are removed from `poly_a`. =#
 function _difference(
     target::TraitTarget{GI.PolygonTrait}, ::Type{T},
     ::GI.PolygonTrait, poly_a,
@@ -114,16 +116,19 @@ function _difference(
     return polys
 end
 
+#= Multipolygon with polygon difference - note that all intersection regions between
+sub-polygons of `multipoly_a` and `poly_b` will be removed from the corresponding
+sub-polygon. Unless specified with `fix_multipoly = false`, `multipolygon_a` will be
+validated using the `UnionIntersectingPolygons` correction. =#
 function _difference(
     target::TraitTarget{GI.PolygonTrait}, ::Type{T},
     ::GI.MultiPolygonTrait, multipoly_a,
     ::GI.PolygonTrait, poly_b;
     fix_multipoly = true, kwargs...,
 ) where T
-    if fix_multipoly
-        multipoly_a = MinimalMultiPolygon()(multipoly_a)
+    if fix_multipoly # Fix multipoly_a to prevent returning an invalid multipolygon
+        multipoly_a = UnionIntersectingPolygons()(multipoly_a)
     end
-    # TODO: Should we fix multipoly_a -> shouldn't happen every time if called with 2 multipoly
     polys = Vector{_get_poly_type(T)}()
     sizehint!(polys, GI.npolygon(multipoly_a))
     for poly_a in GI.getpolygon(multipoly_a)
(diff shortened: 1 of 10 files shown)


src/methods/clipping/difference.jl
```diff
@@ -99,6 +105,65 @@ tracing a_list, else step backwards, where x is the entry/exit status and y is a
...
+
+#= Multipolygon with polygon difference - note that all intersection regions between
+sub-polygons of `multipoly_a` and `poly_b` will be removed from the corresponding
+sub-polygon. Unless specified with `fix_multipoly = false`, `multipolygon_a` will be
+validated using the `UnionIntersectingPolygons` correction. =#
+function _difference(
+    target::TraitTarget{GI.PolygonTrait}, ::Type{T},
+    ::GI.MultiPolygonTrait, multipoly_a,
+    ::GI.PolygonTrait, poly_b;
+    fix_multipoly = true, kwargs...,
```
rafaqz, line 132:
```suggestion
    fix_multipoly = UnionIntersectingPolygons(), kwargs...,
```

Maybe this is more flexible?

src/methods/clipping/difference.jl
```diff
@@ -99,6 +105,65 @@ tracing a_list, else step backwards, where x is the entry/exit status and y is a
...
+) where T
+    if fix_multipoly # Fix multipoly_a to prevent returning an invalid multipolygon
+        multipoly_a = UnionIntersectingPolygons()(multipoly_a)
+        fix_multipoly = false
+    end
+    local polys
+    for (i, poly_b) in enumerate(GI.getpolygon(multipoly_b))
+        polys = difference(i == 1 ? multipoly_a : GI.MultiPolygon(polys), poly_b;
+            target = target, fix_multipoly = fix_multipoly)
```
rafaqz, line 162:
```suggestion
            target, fix_multipoly)
```

src/methods/clipping/difference.jl
```diff
@@ -99,6 +105,65 @@ tracing a_list, else step backwards, where x is the entry/exit status and y is a
...
+function _difference(
+    target::TraitTarget{GI.PolygonTrait}, ::Type{T},
+    ::GI.MultiPolygonTrait, multipoly_a,
+    ::GI.MultiPolygonTrait, multipoly_b;
+    fix_multipoly = true, kwargs...,
+) where T
+    if fix_multipoly # Fix multipoly_a to prevent returning an invalid multipolygon
+        multipoly_a = UnionIntersectingPolygons()(multipoly_a)
+        fix_multipoly = false
+    end
```
rafaqz, line 158:
```suggestion
    ::GI.MultiPolygonTrait, multipoly_b;
    fix_multipoly = UnionIntersectingPolygons(), kwargs...,
) where T
    if !isnothing(fix_multipoly) # Fix multipoly_a to prevent returning an invalid multipolygon
        multipoly_a = fix_multipolyg(multipoly_a)
        fix_multipoly = nothing
    end
```

src/transformations/correction/intersecting_polygons.jl
```diff
@@ -0,0 +1,71 @@
...
+application_level(::UnionIntersectingPolygons) = GI.MultiPolygonTrait
+
+function (::UnionIntersectingPolygons)(::GI.MultiPolygonTrait, multipoly)
+    union_multipoly = if GI.npolygon(multipoly) > 1
+        # Combine any sub-polygons that intersect
+        first_poly = GI.getpolygon(multipoly, 1)
+        exclude_first_poly = GI.MultiPolygon(collect(Iterators.drop(GI.getpolygon(multipoly), 1)))
+        GI.MultiPolygon(union(first_poly, exclude_first_poly; target = GI.PolygonTrait(), fix_multipoly = false))
+    else
+        multipoly
```
rafaqz, line 68: Maybe we want to also return and empty `GI.MultiPolygon` here too for consistency? although empty ones probably don't work currently :sob:

skygering: I think that if there is one or zero (if that was possible) sub-polygons then we should just return the original. There are no fixes to be made regarding "intersecting polygons" given that there cannot be any intersecting polygons.

rafaqz: The issue is it will introduce type instability

skygering: Why? Both branches return multipolygons, right?

skygering: Oh! Is it because one branch might not be a` GI.MultiPolygon`? I should probably convert it with `GO.tuples`.

asinghvi17: That should work!  Would probably still error on empty, though.

skygering:
If we feel good about the code, I will merge!

asinghvi17 (approved):

PR #105, opened 2024-04-12
add prepared geometry wrapper

### LINKED ISSUE #103: Prepared geometry
"Prepared geometry" is basically a regular geometry with an attached cache, which can be a sorted edge list, an Rtree, a monotone chain decomposition, etc.

We should create prepared geometry wrapper structs which hold a geometry + its cache.  It may also contain a field verifying whether the geometry has been validated or not.  (This will require special handling in e.g. `apply`, probably).

Preparations would also require an interface, both to create them, and a trait like interface where we can use multiple libraries' implementations (LibSpatialIndex.jl, SortTileRecursiveTree.jl, NearestNeighbors.jl, etc.)

The wrapper would also have to wrap feature collections, so we can use rtrees defined in things like `.gpkg` files or elsewhere for spatial joins etc.

rafaqz:
Needs tests

Fixes #103


src/prepared.jl
```diff
@@ -0,0 +1,56 @@
...
+GI.isfeaturecollection(::Type{<:Prepared{T}}) where {T} = GI.isfeaturecollection(T)
+
+GI.geometry(x::Prepared) = GI.geometry(parent(x))
+GI.properties(x::Prepared) = GI.properties(parent(x)) 
+
+for f in (:extent, :crs)
+    @eval GI.$f(t::GI.AbstractTrait, x::Prepared) = GI.$f(t, parent(x))
+end
+for f in (:coordnames, :is3d, :ismeasured, :isempty, :coordinates, :getgeom)
+    @eval GI.$f(t::GI.AbstractGeometryTrait, geom::Prepared, args...) = GI.$f(t, geom, args...)
```
asinghvi17, line 23: Should this be 
```suggestion
    @eval GI.$f(t::GI.AbstractGeometryTrait, geom::Prepared, args...) = GI.$f(t, parent(geom), args...)
```
?  Or are you going for something else here

src/prepared.jl
```diff
@@ -0,0 +1,56 @@
+struct Prepared{Pa,Pr}
+    parent::Pa
+    preparations::Pr
+end
+
+Base.parent(x::Prepared) = x.parent
+@inline getprep(p::Prepared, x::Symbol) = getpropery(p.preparations, x)
```
asinghvi17, line 7:
```suggestion
@inline getprep(p::Prepared, x::Symbol) = getproperty(p.preparations, x)
```

asinghvi17:
What are your thoughts about the preparation interface?  We need to be able to take a reliable path for preparations which might have `RTree` but not `MonotoneChain`, for example...

rafaqz:
I'm not sure really!  I'm hoping that will be easier when there is a concrete PR using it. Probably we will need to make some changes to some of this too.

asinghvi17:
Sampling could use this - e.g. `to_edges` could access an `EdgeDecomposition` preparation, and linestrings could have precomputed per-segment arclength, multipolygons can have precomputed area per polygon, triangulation, etc.  

A current list of preparations for #184 would be:
- `EdgeVector`: A vector of edges, per geometry.
- `Triangulation`: A tuple of `(points, faces)`, per polygon.  Multipolygons are larger than this and so don't care.
- `SegmentLengths`: A vector of length per edge, linestring only.
- `SubpolygonArea`: A vector of area per subpolygon, multipolygon only.


I guess each preparation would have a specialized function?  Or we can override e.g. area on multipolygons which have the SubpolygonArea preparation.

PR #108, opened 2024-04-14
Add more descriptive docs + some experiments

asinghvi17:
- Benchmark file to test GO's speed on geometries provided by different packages - needs to be made more portable though
- Code for the `vector-benchmark` plot with the logos and all - needs to be made more portable again
- Experiments:
	- Accurate accumulation - I need to find a good test case for this, but it's basically trying to avoid roundoff error.  It doesn't affect the result much for now, but if dealing with really small then really large numbers it might.
	- Exact predicates: a plot of the orient/incircle predicates for really small numbers.  WIP.
- Paradigms: a more standard-like explanation for apply, applyreduce, and fix.
- Peculiarities: open to rename.  Basically explains some things we do in the source (boolsastypes, etc) in more detail.


docs/src/paradigms.md
```diff
@@ -0,0 +1,31 @@
+# Paradigms
+
+GeometryOps exposes functions like `apply` and `applyreduce`, as well as the `fix` and `prepare` APIs, that represent _paradigms_ of programming, by which we mean the ability to program in a certain way, and in so doing, fit neatly into the tools we've built without needing to re-implement the wheel.
+
+Below, we'll describe some of the foundational paradigms of GeometryOps, and why you should care!
+
+## `apply`
+
+
+The `apply` function allows you to decompose a given collection of geometries down to a certain level, and then operate on it.  In general, its invocation is:
```
rafaqz, line 10:
```suggestion
The `apply` function allows you to decompose a given collection of geometries down to a certain level,  operate on it, and reconstruct it back to the same nested form as the original.  In general, its invocation is:
```

docs/src/peculiarities.md
```diff
@@ -0,0 +1,15 @@
+# Peculiarities
+
+## `_True` and `_False` (or `BoolsAsTypes`)
```
rafaqz, line 3: Internals? We should remove the underscore if we document this, or use the existing Static.jl package.

But probably users can use `true` and `false` and we can keep this hidden.

docs/src/paradigms.md
```diff
@@ -0,0 +1,31 @@
...
+
+```julia
+apply(f, trait::Trait, geom)
+```
+
+Functionally, it's similar to `map` in the way you apply it to geometries - except that you tell it at which level it should stop, by passing a `trait` to it.  
+
+`apply` will start by decomposing the geometry, feature, featurecollection, iterable, or table that you pass to it, and stop when it encounters a geometry for which `GI.trait(geom) isa Trait`.  This encompasses unions of traits especially, but beware that any geometry which is not explicitly handled, and hits `GI.PointTrait`, will cause an error.
+
+`apply` is unlike `map` in that it returns reconstructed geometries, instead of the raw output of the function.  If you want a purely map-like behaviour, like calculating the length of each linestring in your feature collection, then call `map` on the result of `GO.flatten(trait, geom)`, which will decompose each geometry to the given `trait` level and return that decomposition as a flattened vector.
```
rafaqz, line 20: Flatten takes a function

```suggestion
`apply` is unlike `map` in that it returns reconstructed geometries, instead of the raw output of the function.  If you want a purely map-like behaviour, like calculating the length of each linestring in your feature collection, then call  `GO.flatten(f, trait, geom)`, which will decompose each geometry to the given `trait` and apply `f` to it, returning the decomposition as a flattened vector.
```

docs/src/peculiarities.md
```diff
@@ -0,0 +1,15 @@
...
+
+When dispatch can be controlled by the value of a boolean variable, this introduces type instability.  Instead of introducing type instability, we chose to encode our boolean decision variables, like `threaded` and `calc_extent` in `apply`, as types.  This allows the compiler to reason about what will happen, and call the correct compiled method, in a stable way without worrying about 
+
+## What does `apply` return and why?
+
+@rafaqz
```
rafaqz, line 9:
```suggestion
`apply` returns the target geometries returned by `f`, whatever type/package they are from. It is recommended for consistency that `f` returns GeoInterface geometries unless there is a performance/conversion overhead to doing that. Objects that wrapped the target are replaced with GeoInterace.jl wrappers with matching `GeoInterface.trait` to the originals. All non-geointerface iterables become `Array`s.
```

rafaqz: Hmm is that actually true about Tables? probably it should be, but DataFrame should return another DataFrame...

asinghvi17: yes, if the Table type provides a materializer, then we can convert - otherwise it returns a named tuple.

docs/src/peculiarities.md
asinghvi17: In a loop for example it may make sense for users to use `_True` instead of `true`.  But we could definitely switch to Static!

rafaqz: Probably we should try to make sure constprop from true => _True just works, and if not switch to Static.jl

rafaqz: maybe we are just missing som `@inline` and/or `Base.@constprop :aggressive`

docs/src/peculiarities.md
```diff
@@ -0,0 +1,17 @@
...
+
+`apply` returns the target geometries returned by `f`, whatever type/package they are from, but geometries, features or feature collections that wrapped the target are replaced with GeoInterace.jl wrappers with matching `GeoInterface.trait` to the originals. All non-geointerface iterables become `Array`s. Tables.jl compatible tables are converted either back to the original type if a `Tables.materializer` is defined, and if not then returned as generic `NamedTuple` column tables (i.e., a NamedTuple of vectors).
+
+ It is recommended for consistency that `f` returns GeoInterface geometries unless there is a performance/conversion overhead to doing that. 
+
+## Why do you want me to provide a `target` in set operations?
+
+@skygering
+
+Mainly type stability reasons.
```
asinghvi17, line 17:
```suggestion
In polygon set operations like `intersection`, `difference`, and `union`, many different geometry types may be obtained - depending on the relationship between the polygons.  For example, when performing an union on two nonintersecting polygons, one would technically have two disjoint polygons as an output.

We use the `target` keyword to allow the user to control which kinds of geometry they want back.  For example, setting `target` to `PolygonTrait` will cause a vector of polygons to be returned (this is the only currently supported behaviour).  In future, we may implement `MultiPolygonTrait` or `GeometryCollectionTrait` targets which will return a single geometry, as LibGEOS and ArchGDAL do.

This also allows for a lot more type stability - when you ask for polygons, we won't return a geometrycollection with line segments.  Especially in simulation workflows, this is excellent for simplified data processing.
```

asinghvi17:
This should be ready to merge now, in so far as I'm happy with its state.  Experiment pages aren't in the sidebar now, but will be once we figure out the exact predicates and have some nice test cases to show off.  You can access them directly through URL though.

benchmarks/geometry_providers.jl
```diff
@@ -0,0 +1,181 @@
...
+
+PROVIDERS = (ArchGDAL, LibGEOS, GeometryBasics, GI.Wrappers)
+
+
+import Polylabel
+
+water1 = ArchGDAL.fromWKT([readchomp(joinpath(dirname(dirname(pathof(Polylabel))), "test", "data", "water1.wkt")) |> String])
+water2 = ArchGDAL.fromWKT([readchomp(joinpath(dirname(dirname(pathof(Polylabel))), "test", "data", "water2.wkt")) |> String])
+
+water1_1 = GO.fix(water1, corrections = [GO.GEOSCorrection()])
```
skygering, line 25: Should this just be `water1`? Given that `water2` wasn't renamed and `water1_1` isn't used again?

benchmarks/geometry_providers.jl
```diff
@@ -0,0 +1,181 @@
...
+water2 = GO.fix(water2, corrections = [GO.GEOSCorrection()])
+
+using CoordinateTransformations, Rotations
+w1rg = GO.transform(Translation(GO.centroid(water1)) ∘ LinearMap(Makie.rotmatrix2d(π/2)) ∘ Translation((-).(GO.centroid(water1))), water1)
```
skygering, line 29: It would be nice to have some comments on what these different geometries are for.

skygering: Just more comments in general in the benchmark files about what the different sections are doing would be nice as they are quite dense.

docs/src/paradigms.md
```diff
@@ -0,0 +1,31 @@
+# Paradigms
+
+GeometryOps exposes functions like `apply` and `applyreduce`, as well as the `fix` and `prepare` APIs, that represent _paradigms_ of programming, by which we mean the ability to program in a certain way, and in so doing, fit neatly into the tools we've built without needing to re-implement the wheel.
```
skygering, line 3: Nice explanations!

docs/src/peculiarities.md
```diff
@@ -0,0 +1,19 @@
...
+
+## What does `apply` return and why?
+
+`apply` returns the target geometries returned by `f`, whatever type/package they are from, but geometries, features or feature collections that wrapped the target are replaced with GeoInterace.jl wrappers with matching `GeoInterface.trait` to the originals. All non-geointerface iterables become `Array`s. Tables.jl compatible tables are converted either back to the original type if a `Tables.materializer` is defined, and if not then returned as generic `NamedTuple` column tables (i.e., a NamedTuple of vectors).
+
+ It is recommended for consistency that `f` returns GeoInterface geometries unless there is a performance/conversion overhead to doing that. 
+
+## Why do you want me to provide a `target` in set operations?
+
+In polygon set operations like `intersection`, `difference`, and `union`, many different geometry types may be obtained - depending on the relationship between the polygons.  For example, when performing an union on two nonintersecting polygons, one would technically have two disjoint polygons as an output.
```
skygering, line 15: Again very nice explanation!

skygering (approved):

benchmarks/geometry_providers.jl
asinghvi17: I just ran this and realised that GEOS is messing this up...again.  I might just use it raw for now.

benchmarks/geometry_providers.jl
asinghvi17: Will do!  They could definitely use some cleanup as well.

Issue #106, opened 2024-04-12
Try to back most things by staticvectors instead of tuples


asinghvi17:
They have better dispatches on math functions and coordinate transformations.

skygering:
@asinghvi17 Did we decide on `GI.Points` or` StaticVectors`? Is the first step to make an equivalent of the `tuples` function and then just replace the `tuples` calls? Seems like this will have the biggest effects on the functions that actually return new geometries, so basically the clipping and cutting and other transformations. 

I am happy to take a stab at this this weekend with a little clarification.

asinghvi17:
I believe it was `GI.Point(SVector((x, y, z, m)))`

Issue #97, opened 2024-04-05
Let all polygon ops start with a bounding box query


asinghvi17:
If the bounding boxes of the two polygons don't intersect, we can immediately short-circuit the evaluation of any polygon set operation.  Would give up to 1000x speedup in some cases, and of course more for really complex polygons.

rafaqz:
Precalculating the extents may help if we're going to do this multiple time with the same obect

skygering:
@rafaqz Would we then want to store the extent within the `GI` polygons? Would we always precalculate, or would this be a user choice? It would also be good to sort the polygons by extent when validating multipolygons. This would change the types of the returned polygons.

rafaqz:
if you use `tuples` I think you can pass `calc_extent=true` to embed it for all nested geoms

skygering:
See the conversation in #107 for more implementation details.

PR #112, opened 2024-04-16
svector points

skygering:
Switch functions that return geometries to have points of the form `GI.Point{SVector}` rather than a `Tuple`.

This addresses #106 

I did not update `barycentric.jl.` I don't really understand what is going on there and it seems to depend on `GeometryBasics` functions. Do we also want that swapped over? I think it is fine because it returns the lambda values, not an actual object, correct?


asinghvi17:
Hmm, good point on `barycentric` - that code is a little messy but was basically trying to roundtrip the z-values.  In this case I think tuples are probably fine (or at least no better or worse than static vectors).

src/GeometryOps.jl
```diff
@@ -6,16 +6,21 @@ using GeoInterface
...
 
 const GI = GeoInterface
 const GB = GeometryBasics
+const SA = GeometryBasics.StaticArrays
 
 const TuplePoint{T} = Tuple{T, T} where T <: AbstractFloat
-const Edge{T} = Tuple{TuplePoint{T},TuplePoint{T}} where T
+const TupleEdge{T} = Tuple{TuplePoint{T},TuplePoint{T}} where T
+
+const SVPoint{T} = GI.Point{false, false, SA.SVector{2, T}, Nothing} where T <: AbstractFloat
```
asinghvi17, line 20: I'm not sure how to enable `SVPoint{Z, M, T}` - tried to do it the obvious way with:
```julia
const SVPoint{Z, M, T, CRS} = GI.Point{Z, M, SA.SVector{2+Z+M}, T, CRS}
```
but that complains about Z and M being type variables.  Any thoughts on how we could accomplish this would be great :D

src/methods/polygonize.jl
```diff
@@ -67,26 +67,25 @@ If `xs` and `ys` are passed in they are used as the pixel center points.
...
     ## This function uses a lazy map to get contours.  
     contours = Iterators.map(get_contours(A)) do contour
-        poly = map(contour) do xy
+        map(contour) do xy
             x, y = Tuple(xy)
-            Point2f(xs[x], ys[y])
+            _sv_point((xs[x], ys[y]), T)
         end
     end
+    @show typeof(contours)
```
asinghvi17, line 80: should this be removed?

src/transformations/correction/closed_ring.jl
```diff
@@ -48,26 +48,21 @@ struct ClosedRing <: GeometryCorrection end
 
 application_level(::ClosedRing) = GI.PolygonTrait
 
-function (::ClosedRing)(::GI.PolygonTrait, polygon)
-    exterior = _close_linear_ring(GI.getexterior(polygon))
-
+function (::ClosedRing)(::Type{T}, ::GI.PolygonTrait, polygon) where T
```
asinghvi17, line 51: especially here I'm not sure if this is necessary, since we're directly copying the point.  Maybe enforcing standard point datatypes could be another fix, though!

src/transformations/flip.jl
```diff
@@ -14,14 +14,18 @@ original type).
...
+            GI.Point(SA.SVector{4}(GI.y(p), GI.x(p), GI.z(p), GI.m(p)))
+        end
+    elseif _is3d(geom)
+        return apply(PointTrait(), geom; kw...) do p
+            GI.Point(SA.SVector{3}(GI.y(p), GI.x(p), GI.z(p)))
         end
     else
         return apply(PointTrait(), geom; kw...) do p
-            (GI.y(p), GI.x(p))
+            GI.Point(SA.SVector{2}(GI.y(p), GI.x(p)))
```
asinghvi17, line 28: should this be `SVPoint`?

src/transformations/tuples.jl
```diff
@@ -13,7 +13,11 @@ geometries wrapping `Tuple` points.
 $APPLY_KEYWORDS
 """
 function tuples(geom, ::Type{T} = Float64; kw...) where T
-    if _is3d(geom)
+    if _ismeasured(geom)
+        return apply(PointTrait(), geom; kw...) do p
+            (T(GI.x(p)), T(GI.y(p)), T(GI.z(p)), T(GI.m(p)))
```
asinghvi17, line 18: It isn't necessary that a geometry has both Z and M.  This would have to be a little more complex (though I feel like we should define some form of framework for this, even if it's just conceptual and in the docs)

test/methods/centroid.jl
```diff
@@ -104,7 +104,7 @@ end
     ])
     c1, area1 = GO.centroid_and_area(m1)
     c1_from_LG = LG.centroid(m1)
-    @test c1[1] ≈ GI.x(c1_from_LG)
```
asinghvi17, line 107: Ah is this because the points are now `GI.Point`?  We should still allow getindex behaviour on those IMO @rafaqz, just for convenience's sake.

src/GeometryOps.jl
skygering: Yeah... this is also going to be needed for #97  so that we can actually store the extents...

I also don't know how to do it 😆 which is why I hard-coded for now

skygering: Also, the functions that use it only currently work for 2D so it isn't the end of the world right now, but definitely a needed improvement.

src/transformations/correction/closed_ring.jl
skygering: The function was type unstable before since once branch of the LienarRing dispatch used `tuples` and the other didn't. So I just changed it so all branches use `SVPoints`.

src/transformations/flip.jl
skygering: We could probably make something like that work with a basic constructor. I think I didn't do it since we didn't have a generalizable way so I wanted all of the branches to stay the same until we got that figured out...

src/transformations/tuples.jl
skygering: Hmmmm.... It does say in the [GI docs](https://juliageo.org/GeoInterface.jl/stable/reference/api/#GeoInterface.m-Tuple{Any}) that for a point to be measured, it MUST be a 4-element tuple.

test/methods/centroid.jl
skygering: Yep. Would that be implemented in `GI`?

src/transformations/correction/closed_ring.jl
asinghvi17: Ah I meant the addition of `T` here.  Changing to svpoints does make sense.

skygering: Then I would need to do eltype or get rid of the `Float64` assumption or else it will just default anyways. Do we want to do that?

src/transformations/flip.jl
asinghvi17: Ah yeah a function `svpoint{Z, M, [T]}(x, y, [z, [m]])` should work

rafaqz:
I dont 100% get the reason to put SVector in points... its only useful for dispatch and wrapping it loses that.

We could make points `<: StaticVector` ...

asinghvi17:
I guess at this point we would have to make our own point type?  In that case it may as well be backed by a static vector.

rafaqz:
Well GI.Point is my own point type ;)

But the SVector dep might be too heavy for GI. We can dup it here as <: SVector if we really need that.

skygering:
> I dont 100% get the reason to put SVector in points... its only useful for dispatch and wrapping it loses that.
> 
> We could make points `<: StaticVector` ...

I thought that you wanted this @rafaqz ? If not, I honestly also don't like it. I think it is annoying to generate the polygons this way, since `GI` doesn't automatically do the wrapping... I would have rather just done `SVectors` but I thought this is what we talked about in the slack and #106.

Do we just want `SVectors`? Does having the point wrapper do anything that we want it to do?

skygering:
From our conversation on Slack today, it seems like the best course of action is to create our own point type:
```julia
struct Point{N,T,Z,M} <: StaticVector{N,T}
    vals::NTuple{N,T}
end
```
This will allow dispatch off of `Z` and `M`, but still give us the benefits of `StaticVectors`, like indexing.

skygering:
Okay! It is now working! Within `types.jl` I have added the new `SVPoint` type and have 2D, 3D, and 4D constructors to make points of the correct type. Should `Z` and `M` be booleans or `BoolsAsTypes`?

Should there be a more general constructor? We could dispatch off of the type of `NTuple`, but if anything but an `NTuple` is passed in, we will need to do a `_ismeasured` and a `_is3d` check.

This then leads into my next question. Should we try to generalize the `_to_points` and `_to_edges` code? Should we have different functions for `TuplePoints` and `SVPoints`? Because right now they just return 2D tuples. However, that really isn't apparent from the name and requires having the `_sv_points` function. I also tried to collect and consolidate the various functions that turn points and geometries into different point types within the `utils.jl` file.

Finally, I have it so that we are passing `T` around a lot so that the points can be a user-specified type. I think this is nice for type stability... but it doesn't work for the `GEOS` functions, for example, as `LibGEOS` only works with Float64... 

Also, I know I need to add comments on the new types, but I wanted to make sure everything is correct first.

asinghvi17:
Yes, I think a general constructor would be nice.  Generally, it probably should be parametrized to avoid a bunch of if statements. 

Z and M can be booleans as in the GI structs - the values are encoded in the type domain so it's not a problem for dispatch and specialization.  We only really need the BoolsAsTypes when passing around values.

Passing T around should be fine since we do it a bunch...they are probably not applicable to GEOS but when we add GEOSConvert then they could be useful.

I have to read this thoroughly but this is my first impression.

skygering:
I will say - this doesn't even fix my problem with using `transform`. Here, the point type is `SVPoint` but after going through a transform the geometry will have point that are of type `StaticVector`. My big annoyance is that type change, which this doesn't even fix... 

Thoughts on that?

rafaqz:
I think I mentioned in the past that there is not so much use to having a wrapped `SVector`, as it hides all the StaticArrays.jl dispatch anyway. I think the options are:

1. Tuple
2. Point
3. SVector

But not a Point{Svector}

asinghvi17:
The fix for transform would just be changing `StaticArray` to `SVPoint` there, IMO.

src/transformations/transform.jl
```diff
@@ -38,20 +35,21 @@ by the `SVector` point, which is easy using an anonymous function.
...
 ```
 """
-function transform(f, geom; kw...) 
-    if _is3d(geom)
+function transform(f, geom, ::Type{T} = Float64; kw...) where T
+    if _ismeasured(geom)
         return apply(PointTrait(), geom; kw...) do p
-            f(StaticArrays.SVector{3}((GI.x(p), GI.y(p), GI.z(p))))
+            SVPoint_4D(f(StaticArrays.SVector{4}((GI.x(p), GI.y(p), GI.z(p), GI.m(p)))))
```
asinghvi17, line 44:
```suggestion
            SVPoint_4D(f(SVPoint_4D((GI.x(p), GI.y(p), GI.z(p), GI.m(p)))))

**Comment by rafaqz** [comment]
Ok lol I should read it first I misunderstood what SVPoint is.

Do we just need to add some `similar` methods? I've extended static vectors before and it is possible

File: src/transformations/tuples.jl (hunk 5 above)
Comment (asinghvi17) [reply src/transformations/tuples.jl:18]: That's only for tuple and vector points:

> Length 3 Tuple and Vector points

**Comment by skygering** [comment]
Do you have any example code, @rafaqz ? I have never extended an `SVector` so I am a little lost. 

I am also not sure how to make `to_points` and `to_edges` general without doing a conditional check for `is_measured` and `is_3d`. Thoughts on that? Do we just do the check?

**Comment by asinghvi17** [comment]
May as well do the test - it should ideally compile away if that is known at compile time.

**Comment by rafaqz** [comment]
Is3d/ismeasured still won't compile away for some backends where it's a runtime check, so we want the checks as far out as possible... Like nested if blocks and four separate calls to `apply` for each combination will be the most stable

**Comment by rafaqz** [comment]
I'm not sure for SArray in this case, it could be missing constructors as well. Static arrays has so many fallbacks. 

You could use Cthulhu  and see where the type changes, and add that method.

File: src/types.jl [hunk 8]
```diff
@@ -68,6 +68,61 @@ struct _False <: BoolsAsTypes end
...
+_TuplePoint_3D(::Type{T}, ::_True, vals) where T = TuplePoint{3, T}((GI.x(vals), GI.y(vals), GI.m(vals)))
+
+TuplePoint_4D(vals, ::Type{T} = Float64) where T = TuplePoint{4, T}((GI.x(vals), GI.y(vals), GI.z(vals), GI.m(vals)))
+
+#=
+## `SVPoint`
+
+
+=#
+struct SVPoint{N, T, Z, M} <: GeometryBasics.StaticArraysCore.StaticVector{N,T}
```
skygering, line 92: Thoughts on this structure update @asinghvi17 and @rafaqz ?

skygering:
Also all of my tests are passing locally so I am confused as to why they are failing on GitHub.

src/utils.jl
```diff
@@ -66,11 +66,23 @@ Convert any geometry or collection of geometries into a flat
...
+function find_point_constructor(x, ::Type{T})
+    if _ismeasured(x)
+        SVPoint_4D
+    elseif _is3d
+        SVPoint_3D
+    else
+        SVPoint_2D
+    end
+
+end
```
asinghvi17:
```suggestion
function find_point_constructor(x, ::Type{T})
    M = _ismeasured(x)
    Z = _is3d(x)
    N = 2 + M + Z
    return SVPoint{N, T, Z, M}
end
```
This is what we should have, IMO, but it does overlap in purpose with the SVPoint_4D etc constructors defined earlier.  I wonder whether we should simply define a default and a fallback for three-dimensional points, since that's where all the uncertainty is...

asinghvi17:
All the errors are in the table integration tests.  I'll take a look locally - it's possible that there's a `missing` sneaking in somewhere.

src/types.jl
asinghvi17: The struct looks good to me!

src/utils.jl
skygering: What do you mean? I don't understand.

src/types.jl
rafaqz: Yep looks good

PR #135, opened 2024-05-06
add macro to test all implementations

### LINKED ISSUE #134: Test all available geometry sources
We should automate testing across geometry types from:

ArchGDAL.jl
LibGEOS.jl
GeoInterface.jl
GeometryBasics.jl
Meshes.jl

Just using `GI.convert` in a loop over the modules.

And maybe Shapefile,jl and GeoJSON.jl ? These would need write/read to convert, as there is no `GI.convert` for them as its not very useful except for testing like this.

This would make use of GeometryOps.jl much more robust, avoiding errors like #133 

It would also provide some useful tests on the rest of the ecosystems GeoInterface implementations.

rafaqz:
WIP conversion of all tests to use our new `@test_all_implementations` macro to make sure all geometries work everywhere.

@asinghvi17 I added some helpers for nicer testset titles that combine the topic and the module name.

Closes #134

(probably doesn't work currently !)

- [ ] TODO: There is no `GI.convert` for features and feature collections so we cant pass features into the macro. We should add it to GeoInterface with a fallback that just converts the geometries and rewraps them with GI Feature.


test/helpers.jl
```diff
@@ -0,0 +1,36 @@
...
+# Macro to run a block of `code` for muliple modules, 
+# using GeoInterface.convert for each var in `args`
+macro test_all_implementations(args, code)
+    _test_all_implementations_inner("", args, code)
+end
+macro test_all_implementations(title::String, args, code)
+    _test_all_implementations_inner(string(title, " "), args, code)
+end
+
+function _test_all_implementations_inner(args, code)
```
asinghvi17, line 12:
```suggestion
function _test_all_implementations_inner(title, args, code)
```

test/helpers.jl
```diff
@@ -0,0 +1,36 @@
...
+            end
+        end
+    end
+    quote
+        for mod in (GeoInterface, ArchGDAL, GeometryBasics, LibGEOS)
+            @testset "$title : $mod" begin
+                $let_expr
+            end
+        end
+    end
```
asinghvi17, line 35:
```suggestion

    quote
        for mod in (GeoInterface, ArchGDAL, GeometryBasics, LibGEOS)
            @testset "$(isempty(title) ? "" : "$title : " )$mod" begin
                $let_expr
            end
        end
    end

**Review by asinghvi17 — APPROVED** [review]
Looks good, but every test file will have to `include("helper.jl")` because they're wrapped inside `SafeTestsets`.

File: test/helpers.jl [hunk 3]
```diff
@@ -0,0 +1,36 @@
+using Test, GeoInterface, ArchGDAL, GeometryBasics, LibGEOS
+
+# Macro to run a block of `code` for muliple modules,
```
asinghvi17, line 3:
```suggestion
# Macro to run a block of `code` for multiple modules,

File: test/helpers.jl [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+    else # Handle a tuple of variable names
+        quote
+            let ($args = map(g -> GeoInterface.convert(mod, g), $args))
+                $code
+            end
+        end
+    end
+    quote
+        for mod in (GeoInterface, ArchGDAL, GeometryBasics, LibGEOS)
+            @testset "$(isempty(title) ? "" : "$title : " )$mod" begin
```
asinghvi17, line 31:
```suggestion
            @testset "$($(isempty(title) ? "" : "$title : " ))$mod" begin

File: test/helpers.jl [hunk 5]
```diff
@@ -0,0 +1,36 @@
...
+        end
+    else # Handle a tuple of variable names
+        quote
+            let ($args = map(g -> GeoInterface.convert(mod, g), $args))
+                $code
+            end
+        end
+    end
+    quote
+        for mod in (GeoInterface, ArchGDAL, GeometryBasics, LibGEOS)
```
skygering, line 30: Might be nice to pull these out to the top is a `const` list just to make it really clear which ones are tested. That way if other geometries make it into the scene we can easily test.

Also, do we want GeoInterface to be `tuples`? Or do we not care about the point type here?

skygering (commented):
I am glad we have this. In a few places I was trying to make examples of different types of geometries to test this. It is nice that this is automated now.

test/helpers.jl
rafaqz: I'll move it up. Hopefully the point type doesn't matter much...

rafaqz:
Some ecosystem problems to fix here:

[ ] ArchGDAL cant convert to Line https://github.com/JuliaGeo/GeometryOps.jl/actions/runs/9100439649/job/25015291938?pr=135#step:7:301

[ ] And GeometryBasics cant either https://github.com/JuliaGeo/GeometryOps.jl/actions/runs/9100439649/job/25015291938?pr=135#step:7:352

[ ] ArchGDAL cant make empty points with the createpoint constructor https://github.com/JuliaGeo/GeometryOps.jl/actions/runs/9100439649/job/25015291938?pr=135#step:7:558

[ ] LibGEOS cant convert to GeometryCollection https://github.com/JuliaGeo/GeometryOps.jl/actions/runs/9100439649/job/25015291938?pr=135#step:7:672

[ ] GeometryBasics needs to just conert to LineString for LinearRing https://github.com/JuliaGeo/GeometryOps.jl/actions/runs/9100439649/job/25015291938?pr=135#step:7:722

[ ] GeometryBasics has no GI.ncoord on MultiPolygon https://github.com/JuliaGeo/GeometryOps.jl/actions/runs/9100439649/job/25015291938?pr=135#step:7:875

@evetion we finally have a big test suite to throw at everything and a way to do it easily :D

asinghvi17:
I can deal with the GeometryBasics stuff today, will push a change to the CI to use that branch once I have.

asinghvi17:
Looks like the `distance` test failures are because GB and AG polygons hold LineStringTraits, whereas LG and GI polygons hold LinearRingTraits.

rafaqz:
@asinghvi17 do you want to update the barycentric tests with `@test_all_implementations`? none of the geometries have variable names yet to pass to the macro, and you will probably do that naming in a more sensible way than me

asinghvi17:
Sure, will push it up in a bit.

felixcremer:
>  we finally have a big test suite to throw at everything and a way to do it easily :D

Does this mean we should add GeometryOps as reverse dependency test to these other packages?

rafaqz:
Maybe at some stage. We could add an env var to only test one package

test/methods/geom_relations.jl
```diff
@@ -156,14 +156,16 @@ test_pairs = [
...
+        @test_all_implementations (g1, g2) begin
+            if swap_points
+                g1, g2 = g2, g1
+                sg1, sg2 = sg2, sg1
+            end
+            go_val = GO_f(g1, g2)
+            lg_val = LG_f(g1, g2)
+            @test go_val == lg_val
+            go_val != lg_val && println("\n↑ TEST INFO: $sg1 $f_name $sg2 - $sdesc \n\n")
         end
```
asinghvi17, line 168: This is killing test logs - it is literally impossible to read them because of the number of testsets.  There must be a better way to handle this case - maybe using an external loop over the modules instead of an internal one over the geometries?

rafaqz: Inverting the loop wont change the amount of printing as its in the middle already.

I think we instead need to remove that line? Maybe we could make it a testset instead?

asinghvi17: I don't think the printing is an issue because as we fix the test cases, those will naturally fix themselves.  I guess the problem is that each case has an individual testset - we should change that so that each module just runs all the test cases instead.

Maybe in this case we could use an explicit loop instead of the macro?

asinghvi17:
I'm going to make a PR that changes the implementation to be a custom `@test` instead of a custom `@testset`

rafaqz:
@asinghvi17 did you end up working on this? otherwise I can finish it

asinghvi17:
I am currently on the road so haven't got anything done...go for it!   Are you doing this in the a and b list generators?

rafaqz:
not sure you mean by a and b list generators...

rafaqz:
Do you know if there is a macro utility other packages use to walk the AST and replace `$var` with `var`, and make a vars list at the same time? 

It seems like a generic thing someone has written

asinghvi17:
Oops, somehow I confused this for the other issue.  I did a bit of work on this, will push it tomorrow!

asinghvi17:
BenchmarkTools probably has something

rafaqz:
Ok BenchmarkTools `quasiquote!` is what we want:

https://github.com/JuliaCI/BenchmarkTools.jl/blob/cb09e406b84c6a75f257e2b8dbec7fa4f8b8aca5/src/execution.jl#L361-L380

Its small enough to just copy/paste

rafaqz:
This should be good to go. 

I think I've made the printing a bunch shorter. But its still mostly the `@testset_implements` macro as doing this in a `@test_implements` macro doesn't really work - the geoms are used in the shared setup code.

I've also actually used SafeTestsets.jl (we depended on it but didnt use it) and updated everything for that.

One thought I have is we should show wherever we are manually skipping packages - like maybe print a warning about it?

asinghvi17:
I guess I'm not totally clear on where the setup code needs to be package specific?

> I've also actually used SafeTestsets.jl (we depended on it but didnt use it) and updated everything for that.

:+1:

> One thought I have is we should show wherever we are manually skipping packages - like maybe print a warning about it?

What defines "skipping packages"?  One thought I had was to define e.g. `GI.convert(ClockwiseExteriorWindingOrder, geom)` to enable us to test winding order, closed vs unclosed rings, etc.  That can probably go in another PR though.

Similarly I don't think there's infrastructure to `convert` to Shapefile geoms, or GeoJSON geoms.  Are we implicitly skipping those?

test/helpers.jl
```diff
@@ -0,0 +1,119 @@
+module TestHelpers
+
+using Test, GeoInterface, ArchGDAL, GeometryBasics, LibGEOS
+
+export @test_implementations, @testset_implementations
+
+const TEST_MODULES = [GeoInterface, ArchGDAL, GeometryBasics, LibGEOS]
```
asinghvi17, line 7: Should this be configurable by Preferences.jl or an env variable?

asinghvi17:
Overall, aside from those comments, I think we should merge!

asinghvi17 (approved):
I'm sold!

rafaqz:
By skipping I mean manually passing a subset of the defaults - so just getting the setdiff of the passed modules and the standard ones and putting it in `@warn`. But that can be later.

PR #141, opened 2024-05-13
Fix polygonize

rafaqz:
I forgot to PR this months ago.

But it mostly rewites `polygonize` and uses the uglier but more practically useful exact pixel approach rather than having any angles or curves in the output. It also fixes some bugs and lets you define the contour cutoff yourself.

@skygering I added a performance optimization on `coveredby`, because it a huge gain for `polygonize`. 

```julia
# with extent check in `coveredby`
julia> @btime polygonize(>(0.6), rand(100, 100), minpoints=3);
  2.985 ms (5774 allocations: 2.92 MiB)

# without it
julia> @btime polygonize(>(0.6), rand(100, 100), minpoints=3);
  15.703 ms (5625 allocations: 2.93 MiB)
```

~~I'm not sure if that's the way to do it, it should probably go into `_line_polygon_process` so its used for everything. But, there are cases where not intersecting should return `true` and others where it should return `false` depending on the other arguments?~~

Extent checks are done generically for all predicates now

~~Oh and this needs tests!!~~


src/methods/polygonize.jl
```diff
@@ -56,166 +50,180 @@ f
 The implementation follows:
 =# 
 
+abstract type PolygonizeMethod end
+struct Pixels <: PolygonizeMethod end
+struct Angled <: PolygonizeMethod end
```
asinghvi17, line 55: Needs docstrings

src/methods/polygonize.jl
```diff
@@ -56,166 +50,180 @@ f
...
-    newp = pixel + dir_delta[dir]
-    height, width = size(image)
-    if (0 < newp[1] <= height) && (0 < newp[2] <= width)
-        if image[newp] != 0
-            return newp
+polygonize(f::Base.Callable, xs::AbstractVector, ys::AbstractVector, A::AbstractMatrix; kw...) =
+    _polygonize(f, xs, ys, A; kw...)
+
+function _polygonize(f, xs::AbstractVector{T}, ys::AbstractVector{T}, A::AbstractMatrix; 
+    minpoints=10,
```
asinghvi17, line 115: What's the motivation for this default?

src/methods/polygonize.jl
rafaqz: Need deleting :))

src/methods/polygonize.jl
rafaqz: Yeah, good question. I guess it should be zero.

src/methods/geom_relations/geom_geom_processors.jl
```diff
@@ -315,11 +338,25 @@ If out_require is true, the first polygon must have at least one interior point
 If the point is in an "allowed" location and meets all requirments, return true.
 Else, return false.
 =#
-function _polygon_polygon_process(
+@inline function _polygon_polygon_process(poly1, poly2; kw...)
+    skip, returnval = _maybe_skip_disjoint_extents(poly1, poly2; kw...)
+    if skip 
+        return returnval
+    else
+        return _polygon_polygon_process(poly1, poly2; kw...)
```
skygering, line 346: should this be `_inner_polygon_polygon_process`?

src/methods/geom_relations/geom_geom_processors.jl
```diff
@@ -315,11 +338,25 @@ If out_require is true, the first polygon must have at least one interior point
...
+    end
+end
+
+function _inner_polygon_polygon_process(
     poly1, poly2;
     in_allow, on_allow, out_allow,
     in_require, on_require, out_require,
 )
+    skip, returnval = _maybe_skip_disjoint_extents(poly1, poly2;
```
skygering, line 355: shouldn't this be taken care of in the `_polygon_polygon_process` call? 

Also I personally like the `skip && return returnval` format. Maybe use above for the other checks?

src/methods/geom_relations/geom_geom_processors.jl
```diff
@@ -669,4 +707,27 @@ end
...
\ No newline at end of file
+end
+
+# Disjoint extent optimisation: skip work based on geom extent intersection
+# returns Tuple{Bool, Bool} for (skip, returnval)
+@inline function _maybe_skip_disjoint_extents(a, b;
+    in_allow, on_allow, out_allow, 
+    in_require, on_require, out_require,
+    kw...
+)
```
skygering, line 718: I don't think this logic is quite clear and I think it misses the cases where `in_allow = true` and `out_allow = true` but no `in` or `on` points are required. In that case, a `disjoint` could still short circuit and return a `true` and we could be done. I would suggest something like:

```julia
ext_disjoint = Extents.disjoint(GI.extent(a), GI.extent(b))
skip, return_val = if !ext_disjoint
    # can't tell anything about this case
    false, false
elseif out_allow  # && ext_disjoint
    if in_require || on_require
        true, false
    else
        true, true
    end
else  # !out_allow && ext_disjoint
    # points not allowed in exterior, but geoms are disjoint
    true, false
end
return skip, return_val
```

Something to note here is that you shouldn't be able to have a `required` val be `true` without the `allowed` val being `true`. I think that is implicitly checked in my code, but might need to be more explicit. But logically, you can't require a point in a given region if not points are allowed in that region.

src/methods/polygonize.jl
```diff
@@ -57,165 +51,278 @@ The implementation follows:
...
-            break
+function polygonize(f, xs::AbstractVector{T}, ys::AbstractVector{T}, A::AbstractMatrix; 
+    minpoints=0,
+) where T
+    # Define buffers for edges and rings
+    edges = Dict{T,Tuple{Tuple{T,T},Tuple{Bool,Bool}}}()
+    rings = Vector{T}[]
+
+    @assert (length(xs), length(ys)) == size(A)
+
```
skygering, line 149: Maybe less skipped lines here for vertical compression?

src/methods/polygonize.jl
```diff
@@ -57,165 +51,278 @@ The implementation follows:
...
-                    from -= CartesianIndex(0, 1)
-
-                else
-                    nbd += 1
-                    if fji > 1
-                        lnbd = fji
+        currentpoint = firstpoint
+        ring = T[currentpoint, nextpoint]
+        push!(rings, ring)
+        # println()
```
skygering, line 210: remove debugging statements pre-merge

test/methods/bools.jl
```diff
@@ -0,0 +1,156 @@
+using Test, GeometryOps
+import GeoInterface as GI
+import GeometryOps as GO
+
```
skygering, line 4: This file doesn't exist anymore in more current versions of the code, right?

skygering (commented):
Exciting optimizations. I don't think the logic is quite there but let me know what you think about my comments.

test/methods/bools.jl
rafaqz: Ah its a rebase mistake, this was a old PR

src/methods/geom_relations/geom_geom_processors.jl
rafaqz: Ah yes I didn't quite get that required was a subset of allowed. And good to add more cases we can reject, thanks for the edit

src/methods/geom_relations/geom_geom_processors.jl
rafaqz: Yeah, these are just just copy paste errors. 

(I noticed that these functions are big enough that  calling them takes a bit of time on its own, so inlining the check helps a lot - but I messed up the boilerplate)

src/methods/geom_relations/geom_geom_processors.jl
skygering: This is probably why the tests are hanging

src/methods/geom_relations/geom_geom_processors.jl
rafaqz: (And yes `GI.extent` checks for an extent first and calculates if it doesn't exist. `Extents.extent` cant do the calculating part as it doesn't know about geointerface objects, so we added `GI.extent` partly for this role)

src/methods/polygonize.jl
rafaqz: I need a proper error rather than `@assert toolll

asinghvi17:
We should propagate `GI.crs(A)` to the created geometries, that gets this PR 50% of the way there to full Raster support (the other 50% would probably be an extension on DimensionalData.jl).

PR #151, opened 2024-06-10
Geometry creation tutorial

asinghvi17:
This PR adds a geometry creation tutorial plus some other doc improvements that sprang from there.

cc @alex-s-gardner


docs/src/tutorials/creating_geometry.md
```diff
@@ -7,15 +7,17 @@ import GeoFormatTypes as GFT
 import CoordinateTransformations
 import Proj
 using CairoMakie
+using GeoMakie
```
alex-s-gardner, line 10: Note GeoMakie needs to be loaded for any plotting to work, so I've moved all needed packages to the start

asinghvi17: Ah, that's because GeoInterfaceMakie needs to be loaded as well.  GeoInterface can't actually do that since it's a dependency of Makie, any other package geometry would not require this.

asinghvi17:
https://juliageo.org/GeometryOps.jl/previews/PR151/tutorials/creating_geometry

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,251 @@
...
+The first thing we need to do is decide which Coordinate Reference System (CRS) we will be working in. Here, we start with the most common geographic CRS (i.e. coordiantes of latitude and longitude), [WGS84](https://epsg.io/4326).
+
+````@example creating_geometry
+crs = GFT.EPSG(4326)
+````
+
+Let's start by making a single `Point` with CRS info included.
+
+````@example creating_geometry
+point = GI.Point(0, 0; crs)
```
rafaqz, line 23: Would really like to discourage people from doing this with points... Crs on the wrapper is always enough

alex-s-gardner: I think this should be addressed with a Note and not a change. As we ease users into how to great geospatial geometries a point with crs is the simplest geometry that can be created. But we do need a note on how the CRS is interpreted and that is only needs to be included in the highest level wrapper.

rafaqz: A point without crs is the simplest!!

The idea that points need CRS will stick if it's the first learning experience.

rafaqz: It has terrible consequences, like you can't `reinterpret` the vector as tuples, and you use 1.5 - 2x the memory.

alex-s-gardner: If I wanted to plot a single point on a map, and have the CRS info interpreted, is this not how I would construct it?

rafaqz: But that's an edge case that basically no-one does.

Creating a Point to put in something is the 99% use case, and we should teach those, not the 1%

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,251 @@
...
+y = [-5, 5, 5, -5]
+multipoint = GI.MultiPoint(GI.Point.(zip(x, y); crs))
+plot!(ax, multipoint, marker = '☁', markersize = 30)
+fig
+````
+
+Let's create a line between two points.
+
+````@example creating_geometry
+p1 = GI.Point.(-5, 0; crs)
```
rafaqz, line 55: Like now these have crs too... Even though they're going in the Line strong.

alex-s-gardner: OK, ok... we'll remove the crs. It seems like higher-level constructors should strip lower level CRS info.. and I'm guessing there is a good reason that this wasn't implemented (reallocation?).

rafaqz: That would require a second allocation (bad). Also confusing Because GI geoms just wrap whatever they are given as-is, including points from other packages that they don't know how to modify.

(Im making this point like this because I have thought about it a lot, like while writing the code for GI geoms...)

alex-s-gardner:
I tried to add an example where for plotting a polygon with a different CRS (EPSG:32611) on the same map as the tutorial by was unable to get it to work. I think this would be useful for showing how different `CRS` are interpreted

alex-s-gardner:
> I tried to add an example where for plotting a polygon with a different CRS (EPSG:32611) on the same map as the tutorial by was unable to get it to work. I think this would be useful for showing how different `CRS` are interpreted

This was my attempt:
```
# OK, now what if we want to plot geomitires that are in a diffent Coordinate Reference System (CRS?)

# lets create a polygon in the WGS 84 / UTM zone 10N coodinate system [EPSG:32610](https://epsg.io/32610)
crs2 = GFT.EPSG(4326)

# create new points 
r = 200;
k = 1000;
ϴ = 0:0.01:2pi
x = r .* (k + 1) .* cos.(ϴ) .- r .* cos.((k + 1) .* ϴ)
y = r .* (k + 1) .* sin.(ϴ) .- r .* sin.((k + 1) .* ϴ)

# now a `LinearRing`
ring3 = GI.LinearRing(Point.(zip(x, y)))

# now lets make the LineRing into a Polygon with new `CRS``
polygon3 = GI.Polygon([ring1]; crs=crs2)

# create a new `GeoAccess` with a new source `CRS1`
ga = GeoAxis(
    fig[1, 1];
    source=crs2,
    dest=dest,
    xticklabelsvisible=false,
    yticklabelsvisible=false,
);
```

asinghvi17:
I don't see a definition for dest here - but there are a few examples in the GeoMakie docs that should help

alex-s-gardner:
> but there are a few examples in the GeoMakie docs that should help
I can't seem to find any example with "linked axes" for plotting multiple source `CRS` data on the same figure. Maybe this is not possible yet.

rafaqz:
Ok now I read it all this is awesome! Really is something we need.

Thanks @asinghvi17 I should have written this a year ago

asinghvi17:
hmm, the whole idea is to set the `source` kwarg.  Something like:
```julia
using CairoMakie, GeoMakie
import GeoFormatTypes as GFT
using Interpolations # for Streamplot on gridded data if necessary

f(x::Point2{T}) where T = Point2{T}(
    10 * (2 * cos(2 * deg2rad(x[1]) + 3 * deg2rad(x[2] + 30)) ^ 2),
    20 * cos(6 * deg2rad(x[1]))
)

pole_longitude=177.5
pole_latitude=37.5
fig, ax, plt = streamplot(
    f, 311.9..391.1, -23.6..24.8;
    arrow_size = 6,
    source = "+proj=ob_tran +o_proj=latlon +o_lon_p=0 +o_lat_p=$(pole_latitude) +lon_0=$(180+pole_longitude) +to_meter=$(deg2rad(1) * 6378137.0)",
    axis = (;
        type = GeoAxis,
        title = "Streamplot",
    )
)
lp = lines!(ax, GeoMakie.coastlines(); source = GFT.EPSG(4326), linewidth = 0.5, color = :black, xautolimits = false, yautolimits = false)
translate!(lp, 0, 0, -1)
fig
```

alex-s-gardner:
> whole idea is to set the `source` kwarg

Got it, thanks!

asinghvi17:
I guess that wasn't made clear in the GeoMakie docs.  I'll add a doc page about GeoAxis, and I guess the "multiple CRS" example doesn't show it off very well :D

alex-s-gardner:
Are there plans to have the source `crs` read directly from the geometries? It seems a bit silly that we keep associating `crs` info to the geometries but then they're not used by GeoMakie.

asinghvi17:
I have a branch with some code that does that, unfortunately it causes a stack overflow somehow.  I have to figure that out first but it's very possible.

asinghvi17:
IMO this is pretty good to go from my end, what do you think @alex-s-gardner?

alex-s-gardner:
How can I see the latest preview of the docs like you posted earlier?

asinghvi17:
It's the same link, the deployment can take a couple of minutes sometimes though

alex-s-gardner:
None of the GeoMakie plots seem to render, at least in what I'm looking at

asinghvi17:
Huh, maybe clear your browser cache or reload the page a bunch of times?

https://juliageo.org/GeometryOps.jl/previews/PR151/tutorials/creating_geometry

I just dragged this image from the webpage, it renders with all plots for me
xitdhys DhKQgdJf

asinghvi17:
(there is an issue with the Makie theme leaking, I'll try to fix that)

alex-s-gardner:
> renders with all plots for me

Do you see this image:
Screenshot 2024-06-11 at 7 40 59 PM

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,338 @@
...
+````@example creating_geometry
+xoffset = 0.
+yoffset = 50.
+f = CoordinateTransformations.Translation(xoffset, yoffset)
+polygon1 = GO.transform(f, polygon1)
+plot!(polygon1)
+fig
+````
+
+== Direct manipulation
```
alex-s-gardner, line 114: Can you add some guidance on why one would choose Direct manipulation over CoordinateTransformations ? Having both is a bit confusing

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,338 @@
...
+````@example creating_geometry
+xoffset = 50.
+yoffset = 0.
+f = CoordinateTransformations.Translation(xoffset, yoffset)
+polygon1 = GO.transform(f, polygon1)
+plot!(polygon1)
+fig
+````
+
+== Direct manipulation
```
alex-s-gardner, line 165: Same comment here.

asinghvi17:
I don't see that picture, good point - looking at the logs it seems that GeoMakie's master branch was not being added.  I added that explicitly in an extra step in the CI, it should work at the next deployment.

docs/src/tutorials/creating_geometry.md
asinghvi17: The main reason I wanted to show this is to show users that you can do this directly and don't have to rely on CoordinateTransformations.  

CoordinateTransformations as a package is also only particularly good for affine transformations.  So if you wanted to transform the coordinates using your own polar transform or something, it may not be immediately clear how.  Then again, at that point you've probably read the docs...

If it's too confusing we can remove it, but IMO it's nice to at least show the option.

alex-s-gardner: Given that this is suppose to be a gentle introduction, and is in now way verbose, I'm inclined to omit  `Direct manipulation` as it uses a `do` block which I find the absolutely worst syntax for intuitive understanding... but if you object let's publish

alex-s-gardner:
@asinghvi17  This looks GTO to me.

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+lines = GI.LineString(GI.Point.(zip(x,y)); crs)
+plot!(ax, lines; linewidth = 5)
+fig
+````
+
+We can also create a single `LinearRing` trait, the building block of a polygon.
+A `LinearRing` is simply a `LineString` with the same beginning and endpoint,
+i.e., an arbitrary closed shape composed of point pairs.
+
+A `LinearRing` is composed of a series of points listed in clockwise order (i.e., winding order).
```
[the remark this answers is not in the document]
asinghvi17: That's true, but I suspect polygons will need the winding order.  Maybe we move the explanation to the polygon section instead?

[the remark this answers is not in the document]
rafaqz: Where will they need it though?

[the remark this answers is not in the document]
rafaqz: https://gis.stackexchange.com/questions/119150/order-of-polygon-vertices-in-general-gis-clockwise-or-counterclockwise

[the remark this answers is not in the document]
rafaqz: There is nowhere in GeoInterface that fixes the winding order of shapefiles currently. Maybe it should. And some other formats allow both ways.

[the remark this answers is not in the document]
asinghvi17: Right but do our algorithms actually work with arbitrary winding order?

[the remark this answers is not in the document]
rafaqz: I havent written anything winding dependent. @skygering have you?

I guess we need to extend the tests to generate Shapefile.jl objects with the reveresed windings.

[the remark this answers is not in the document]
asinghvi17: The signed area algorithm might be winding dependent, I have to check.  I remember some issues with the wrong winding order though.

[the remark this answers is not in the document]
skygering: I have tried to write them so they work with arbitrary winding order. Even if I didn't, I think we have to since we accept polygons with arbitrary winding order. If not, then we need to add another "fix" function that changes the winding order. But overall, I agree with @rafaqz

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+import GeoFormatTypes as GFT
+import CoordinateTransformations
+import Proj
+using CairoMakie
+using GeoMakie
+using GeoJSON
+Makie.set_theme!(Makie.MAKIE_DEFAULT_THEME) # hide
+````
+
+The first thing we need to do is decide which [Coordinate Reference System (CRS)](https://en.wikipedia.org/wiki/Spatial_reference_system) we will be working in. Here, we start with the most common geographic CRS (i.e. coordinates of latitude and longitude), [WGS84](https://epsg.io/4326).
```
skygering, line 15: At this point in the tutorial, I still don't quite understand the point of `CRS`. After introducing it, you show in the first few plots that you don't really need it to get the correct points. I would either add a note that we will return to the importance of CRS or I would introduce it later.

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+ring1 = GI.LinearRing(GI.getpoint(lines))
+````
+
+Now, let's make the `LinearRing` into a `Polygon`.
+
+````@example creating_geometry
+polygon1 = GI.Polygon([ring1]; crs)
+````
+
+Now, we can use GeometryOperations and CoordinateTransformations to shift `polygon1`
```
skygering, line 98: Should be `GeometryOps`

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+    Hopefully, these are details that you'll never have to deal with.  But it is good to know.
+
+````@example creating_geometry
+hole = GI.LinearRing(GI.getpoint(multipoint))
+polygon1 = GI.Polygon([ring1, hole]; crs)
+````
+
+Shift `polygon1` to the right, to avoid plotting over our earlier results.
+
+````@example creating_geometry
```
skygering, line 130: When you plot the polygon with the hole, you still call it `polygon1` in the sample code, despite calling the polygon without a hole also `polygon1`. I think I would call this `polygon2` to show it is distinct and then shift the numbering of the polygons below.

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+fig
+````
+
+Great, now we can make `Points`, `MultiPoints`, `Lines`, `LineStrings`, `Polygons` (with holes), and `MultiPolygons` and modify them using [`CoordinateTransformations`] and [`GeometryOps`].
+
+But where does the `crs` information come in? To show this, we need to use `GeoMakie` that can interpret the `crs` information that we've included with our geometries.
+
+Now specify the source and destination projections for our map. Remember that the very first thing we did was select our source coordinate system.
+
+````@example creating_geometry
```
skygering, line 167: I would explain what `source` and `dest` mean in this context right here. As someone who has never used this, I still don't really know what they mean (although you do have a note on the meaning of `source` way down the page).

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+source = crs;
+dest = "+proj=natearth2" #see [https://proj.org/en/9.4/operations/projections/natearth2.html]
+````
+
+Open the Natural Earth continental outlines that are bundled with GeoMakie, and also available from https://www.naturalearthdata.com/.
+
+````@example creating_geometry
+land_path = GeoMakie.assetpath("ne_110m_land.geojson")
```
skygering, line 175: What is this particular land path? I am left wondering what `ne_110m_land` means.

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+````
+
+Plot `land` for context.
+
+````@example creating_geometry
+poly!(ga, land_geo, color=:black)
+fig
+````
+
+Now let's plot a `Polygon` like before, but now on coordinate reference system (CRS) that is different from our data
```
skygering, line 205: You say here that the CRS is different from our data.. but then on line 212 you say we will "plot geometries on the same figure that are in a different Coordinate Reference System (CRS)". So how are these different?

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+import GeoParquet
+
+
+df = DataFrame(geometry=[polygon1, polygon2])
+````
+
+Now let's add a couple of attributes to the geometries
+
+````@example creating_geometry
+df[!,:id] = ["a", "b"]
```
skygering, line 267: What do the `!` do?

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+````
+
+now let's save as a `GeoParquet`
+
+````@example creating_geometry
+fn = "shapes.parquet"
+GeoParquet.write(fn, df, (:geometry,))
+````
+
+Finally, if there's no Julia-native package that can write data to your desired format (e.g. `.gpkg`, `.gml`, etc), you can use [`GeoDataFrames`](https://github.com/evetion/GeoDataFrames.jl).
```
skygering, line 293: I would give a one-sentence comment on why you might pick each of the above packages for writing if you are going to bring them all up.

skygering (commented):
Love it!! I think the CRS part could use a little more polishing but I am happy we have such thorough documentation!

docs/src/tutorials/creating_geometry.md
asinghvi17: That's a good point.  I think the tutorial started as "how to create a geometry with CRS" then evolved to "how to create geometry".  

@alex-s-gardner I guess we could move this a bit further down?

docs/src/tutorials/creating_geometry.md
```diff
@@ -0,0 +1,303 @@
...
+
+Polygons can contain "holes". The first `LinearRing` in a polygon is the exterior, and all 
+subsequent `LinearRing`s are treated as holes in the leading `LinearRing`.
+
+`GeoInterface` offers the `GI.getexterior(poly)` and `GI.gethole(poly)` methods to get the 
+exterior ring and an iterable of holes, respectively.
+
+!!! note
+    Some packages always consider the secondary `LinearRings` holes, others look at the winding
+    order, where the polygons are filled inward if they have a [clockwise](@ref GeometryOps.isclockwise) winding order and
```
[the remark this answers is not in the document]
asinghvi17: Will push a fix.

docs/src/tutorials/creating_geometry.md
asinghvi17: For GeoMakie, source is the CRS of the input and dest is the CRS you want to visualize in.  That could be made a bit clearer, I think - is there a particular verbiage you're thinking of?

docs/src/tutorials/creating_geometry.md
alex-s-gardner: Ya, the crs stuff is a bit strange. I think as the ecosystem matures it will make more sense. What if introduce `crs` right before we make our first GeoAxis?

docs/src/tutorials/creating_geometry.md
asinghvi17: This is the natural earth 110m land polygons.  We could download this straight from NaturalEarth and add a comment about GeoMakie also shipping it.  

An explanation might look like:
```md
Download and open the [Natural Earth](https://www.naturalearthdata.com) global land polygons at 110m resolution.  This will serve as our background image to add context to the map.

```

docs/src/tutorials/creating_geometry.md
asinghvi17: That's DataFrames.jl syntax to mutate a column.  Pretty standard over there, we can link to a description though.

docs/src/tutorials/creating_geometry.md
asinghvi17: That sounds good to me, it's right next to its use there as well.

docs/src/tutorials/creating_geometry.md
skygering: I think even that but of clarification is perfect.

docs/src/tutorials/creating_geometry.md
skygering: That sounds great to me!

alex-s-gardner:
@asinghvi17 is there a function for adding crs info to an existing geometry ... I was looking for something line `setcrs`

asinghvi17:
I don't think we have anything like that yet.  Probably better to just create a new geometry with the CRS.

alex-s-gardner:
Stupid codespace server crashed when I was about to submit my modification... off to meetings now. I will get you something in a few hours

rafaqz:
It would be nice to be able to rebuild any geom with crs without needing to use the right constructor.

alex-s-gardner:
Is it by design that GeoMakie unable to plot a `Line` but it can plot a `LineString`?

asinghvi17:
I'm pretty sure that there was just no convert_arguments added for that on the Makie end.

alex-s-gardner:
> add `setcrs` like Rasters

`setcrs` of `addcrs` is more intuitive than `rebuild`... I don't think most people will intuitively think of adding a `crs` as rebuilding a geometry

docs/src/tutorials/creating_geometry.md
alex-s-gardner: fixed

docs/src/tutorials/creating_geometry.md
alex-s-gardner: fixed

docs/src/tutorials/creating_geometry.md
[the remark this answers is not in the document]
alex-s-gardner: This needs to be carefully traced... winding order has been a pain in my but for years.. if JuliaGeo solves this than all the better but if we don't track carefully it could introduce bugs and compatibility issues... I believe shapefiles use  winding order. For this tutorial I've removed discussion.

alex-s-gardner:
> How would you treat a vector of lines? Would it be continuous (as a LineString) or discontinuous (as individual linesegments)?

This is a question for @rafaqz

alex-s-gardner:
Moving some additional thoughts from the PR back over to Slack

rafaqz:
I imagine given a vector of lines we should plot each one separately.

asinghvi17:
I just pushed some `Line` conversion methods to my GeoMakie PR.  It's a bit inefficient, but should be fine.

docs/src/tutorials/creating_geometry.md
```diff
@@ -33,33 +34,30 @@ fig, ax, plt = plot(point)
...
-multipoint = GI.MultiPoint(GI.Point.(zip(x, y)); crs)
+x = [-5, -5, 5, 5];
+y = [-5, 5, 5, -5];
+multipoint = GI.MultiPoint(GI.Point.(zip(x, y)));
 plot!(ax, multipoint; marker = '☁', markersize = 30)
 fig
 ````
 
-Let's create a line between two points.
+Let's create a `Line` between two points.
```
asinghvi17, line 54:
```suggestion
Let's create a `LineString` connecting two points.

File: docs/src/tutorials/creating_geometry.md [hunk 17]
```diff
@@ -51,7 +51,7 @@ plot!(ax, multipoint; marker = '☁', markersize = 30)
 fig
 ````
 
-Let's create a `Line` between two points.
+Let's create a `LineString` connecting two points.
```
alex-s-gardner, line 57: So here I thought it would be most intuitive to start with a `Line` between two points then to move next to `LineString`, like we did from `Point` to `MuliPoint`... the only reason the demo code used `LineString` is because `Line` would not plot. If a `Line` will not plot in GeoMakie then we could change the code not the text

alex-s-gardner: Let's leave it how you have it here and we can update later once `Line` can be plotted on a released version of Makie

alex-s-gardner (commented):
Let's merge this PR

docs/src/tutorials/creating_geometry.md
asinghvi17: Line is pretty rare and a lot of C libraries don't actually support it.   IMO it's probably best not to advertise it in a user tutorial - there are a lot of sharp edges there.

docs/src/tutorials/creating_geometry.md
```diff
@@ -208,22 +195,22 @@ poly!(ga, land_geo, color=:black)
...
-This is our new source `CRS`: the UTM (Universal Transverse Mercator) zone 10N [EPSG:32610](https://epsg.io/32610).  It's a type of [transverse Mercator projection](https://en.wikipedia.org/wiki/Transverse_Mercator) (a cylindrical projection) that covers the west coast of the U.S. and some regions in Canada.
+To show how to do this let's create a geometry with coordinates in UTM (Universal Transverse Mercator) zone 10N [EPSG:32610](https://epsg.io/32610).
 
 ````@example creating_geometry
-crs2 = GFT.EPSG(32610)
+source_crs2 = GFT.EPSG(32610)
 ````
 
-Create a polygon in the new coordinate system (we're working in meters now, not lat/lon)
+Create a polygon (we're working in meters now, not latitude and longitude)
```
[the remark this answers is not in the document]
asinghvi17: I've written a quick workaround with `DisplayAs` (this is the Documenter-recommended solution).

asinghvi17:
One last thing - I was going through this and figured we needed some anchor points, so added headers.  I will merge for now, but if the headers seem unreasonable we can always change them!

Thanks a lot for the effort @alex-s-gardner (and @JuliaGeo/geometryops for the reviews :D )

Issue #148, opened 2024-06-08
Curried predicate functions


asinghvi17:
Basically defining e.g. `GO.intersects(x) = Base.Fix1(GO.intersects, x)`.  Turns out to be quite useful when performing dataframe ops.

luntan-maker:
I wouldn't mind taking a shot at this, do you do an assignee system? Also just the usual PR?

asinghvi17:
No particular assignee system I think, but a regular PR would be fine - I'll hold off from implementing this then.  Great to see people interested in the project!

rafaqz:
Thanks! No need for a system at this stage, probably whoever is free will review your PR pretty quickly. 

You can also make a draft PR and ask questions at that stage too if there are and problems.

luntan-maker:
Thank you for the warm welcome! I went ahead and made a PR, I look forward working with you.

asinghvi17:
Implemented in #304

PR #152, opened 2024-06-10
Currying functions for performance gain

luntan-maker:
I went through all the files in `src/*`, and curried functions that were repeatedly used within a while or for loop outside of an if statement. I've tried to have some form of literate programming by maintaining the naming convention and appending `_partial`. I did not add comments, since it seemed somewhat straightforward what was going on. I would be more than happy to add comments if you will tell me what you would like me to explain. This should address #148 .

### DIFF
--- a/benchmarks/benchmarks.jl
+++ b/benchmarks/benchmarks.jl
@@ -54,7 +54,7 @@ n_points_values = round.(Int, exp10.(LinRange(log10(10), log10(100_000), 5)))
     circle_area_suite["LibGEOS"][n_points] = @be LG.area($lg_circle) seconds=1
 end
 
-plot_trials(circle_area_suite)
+# plot_trials(circle_area_suite)
 
 
 # ## Vancouver watershed benchmarks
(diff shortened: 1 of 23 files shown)


rafaqz:
Ok, so I feel there may have been some misunderstanding somewhere!

The idea was to add these methods everywhere for users to call themselves:
```julia
GO.intersects(x) = Base.Fix1(GO.intersects, x)
```

As anshul demonstrated. This gives us a "curried" `Fix1` objects to use in `map` and similar functions for the convenience and nice syntax. So we could do `map(intersects(mygeom), othergeoms)` to get a vector of Bool, rather than always having to do `map(g -> intersects(g, mygeom), othergeoms)`. Does that make sense?

Currying anonymous functions explicitly like you have here can sometimes give a performance gain where it can help the compiler (for pretty arcane reasons), but I'm not sure it will be measureable in this case.

luntan-maker:
I think I understand what you are saying. 
Just to be sure though. You want to use `Base.Fix1` and `Base.Fix2` to clean up and make the code more readable/semantic.
This leaves us with two options; build on top of what I already have, or remove it. Which would you prefer?

rafaqz:
Hmm... not really! 

Essentially I'm talking about a different PR to what you have! It would just define methods like this:

```juila
GO.intersects(x) = Base.Fix1(GO.intersects, x)
GO.disjoint(x) = Base.Fix1(GO.disjoint, x)
# etc etc
```

For all of the predicates in the `geom_relations` folder. It would also add tests for all of these in the related test files.

luntan-maker:
I didn’t get what you are saying from what they were saying. Also, I completely missed the part where you said draft PR. So, with your okay, I plan to do the following

1. 	Close this PR
2. 	Clone again or revert current repository
3. 	Open a draft PR
4. 	Start working on what I believe you want

Sound good?

rafaqz:
First maybe write out in your own words what you think we want! Just to be sure there is no communication issue this time

luntan-maker:
Interesting. Thank you for that response. That gives me clarity. I feel this issue is not a good fit for me. You'll need to get someone else to work on this for you. Thank you for the opportunity. Best of luck to you.

rafaqz:
No worries

PR #154, opened 2024-06-12
Small clipping bug

skygering:
Small clipping bug fix. 

Are we okay with a patch bump?

### DIFF
--- a/src/methods/clipping/difference.jl
+++ b/src/methods/clipping/difference.jl
@@ -161,7 +161,16 @@ function _difference(
     end
     local polys
     for (i, poly_b) in enumerate(GI.getpolygon(multipoly_b))
-        polys = difference(i == 1 ? multipoly_a : GI.MultiPolygon(polys), poly_b; target, fix_multipoly)
+        #= Removing intersections of `multipoly_a`` with pieces of `multipoly_b`` - as
+        pieces of `multipolygon_a`` are removed, continue to take difference with new shape
+        `polys` =#
+        polys = if i == 1
+            difference(multipoly_a, poly_b; target, fix_multipoly)
+        else
+            difference(GI.MultiPolygon(polys), poly_b; target, fix_multipoly)
+        end
+        #= One multipoly_a has been completly covered (and thus removed) there is no need to
+        continue taking the difference =#
         isempty(polys) && break
     end
     return polys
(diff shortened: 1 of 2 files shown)


src/methods/clipping/difference.jl
```diff
@@ -162,6 +162,7 @@ function _difference(
     local polys
     for (i, poly_b) in enumerate(GI.getpolygon(multipoly_b))
         polys = difference(i == 1 ? multipoly_a : GI.MultiPolygon(polys), poly_b; target, fix_multipoly)
+        isempty(polys) && break
```
asinghvi17, line 165: Doesn't this fail if multiply A only intersects the 2nd polygon in multiply B?

(Multiply = multi poly, on mobile)

asinghvi17:
Patch bump is fine

src/methods/clipping/difference.jl
asinghvi17: Specifically, considering the case where `multipoly_a` is really just `MultiPolygon(first(GI.getgeom(multipoly_b)))`.  In this case, polys will be empty at i=1, but that short circuiting will ignore the other polygons in poly_b, which are different from multipoly_a.

src/methods/clipping/difference.jl
```diff
@@ -162,6 +162,7 @@ function _difference(
     local polys
     for (i, poly_b) in enumerate(GI.getpolygon(multipoly_b))
         polys = difference(i == 1 ? multipoly_a : GI.MultiPolygon(polys), poly_b; target, fix_multipoly)
```
rafaqz, line 164: We should split out this conditional into an is/else, this logic is a bit hard to follow.

rafaqz: Maybe with a few comments... Given the bug and anshuls comment

src/methods/clipping/difference.jl
skygering: That doesn't matter because we are taking the difference from multipolygon `a`. So once multipolygon `a` is completely covered by any pieces of multipolgon `b`, we have no piece of multipolygon `a` left to take the difference from with the remaining `b` pieces.

skygering: This is a unique short circuit that applies to `difference`. I 100% agree this type of logic doesn't work with `intersection` or `union`.

src/methods/clipping/difference.jl
skygering: Updated!

src/methods/clipping/difference.jl
asinghvi17: Gotcha - that  makes sense.

asinghvi17 (approved):

PR #157, opened 2024-06-13
[WIP] Experimental adjacency graph method + perimeter

asinghvi17:
Currently in `docs/src/experiments/adjacency_graph.jl`.

This PR also adds a perimeter method that can have several algorithms - linear, geodesic, and thumb.  That is still WIP.

The adjacency matrix calculation "works" - it's unidirectional from source to target.

## Cool pictures

download-7

https://github.com/JuliaGeo/GeometryOps.jl/assets/32143268/1933b065-d9f1-4c87-affd-08801df78c6b

https://github.com/JuliaGeo/GeometryOps.jl/assets/32143268/00f99807-28ab-45c4-ab50-a8b5aea4b0d0


asinghvi17:
cc @00krishna

asinghvi17:
A visualization like [the NYT Olympic medal visualization](https://archive.nytimes.com/www.nytimes.com/interactive/2008/08/04/sports/olympics/20080804_MEDALCOUNT_MAP.html?auth=login-smartlock&login=smartlock) would be pretty easy to implement using this method....

asinghvi17:
@00krishna would you want this as like a 2-stage thing, where you can opt to either get the whole row for source and target or just the geometries?

00krishna:
Hey @asinghvi17 sorry for not getting back to you soon, just been kinda swamped.
I have been playing around with this code to get a feel for it. Hmm interesting idea 
about the 1 or 2 stage thing. Yeah the 2 stage things sounds better, in the sense that the 
user can get either the source-target edgelist, and/or the geometries. Sometimes the user just wants the edgelist to work with in a different program, while in other cases the user wants the geometries to plot. So having that option is nice. 

As a caveat, of course there are ways to get the edgelist from the geometries, but users who are less experience will probably just end up posting lots of questions on how to do that. So easier to save ourselves some trouble :).

asinghvi17:
True, I guess there could be multiple entry points to the function then.  Will take a look at how that could be done potentially.

PR #160, opened 2024-06-13
Add a convex hull algorithm

asinghvi17:
Via DelaunayTriangulation.jl.

Needs to be updated to latest main as well as some tests + a GEOS algorithm.

### DIFF
--- a/src/GeometryOps.jl
+++ b/src/GeometryOps.jl
@@ -4,16 +4,17 @@ module GeometryOps
 
 using GeoInterface
 using GeometryBasics
-import Tables
 using LinearAlgebra, Statistics
+
 import GeometryBasics.StaticArrays
+import DelaunayTriangulation # for convex hull and triangulation
 import ExactPredicates
 import Base.@kwdef
-
-using GeoInterface.Extents: Extents
+import GeoInterface.Extents: Extents
 
 const GI = GeoInterface
 const GB = GeometryBasics
+const DelTri = DelaunayTriangulation
 
 const TuplePoint{T} = Tuple{T, T} where T <: AbstractFloat
 const Edge{T} = Tuple{TuplePoint{T},TuplePoint{T}} where T
@@ -28,6 +29,7 @@ include("methods/area.jl")
 include("methods/barycentric.jl")
 include("methods/buffer.jl")
 include("methods/centroid.jl")
+include("methods/convex_hull.jl")
 include("methods/distance.jl")
 include("methods/equals.jl")
 include("methods/clipping/predicates.jl")
--- a/src/GeometryOps.jl
+++ b/src/GeometryOps.jl
@@ -15,7 +15,6 @@ import GeoInterface.Extents: Extents
 
 const GI = GeoInterface
 const GB = GeometryBasics
-const DelTri = DelaunayTriangulation
 
 const TuplePoint{T} = Tuple{T, T} where T <: AbstractFloat
 const Edge{T} = Tuple{TuplePoint{T},TuplePoint{T}} where T
--- a/src/GeometryOps.jl
+++ b/src/GeometryOps.jl
@@ -6,6 +6,7 @@ using GeoInterface
 using GeometryBasics
 using LinearAlgebra, Statistics
 
+import Tables
 import GeometryBasics.StaticArrays
 import DelaunayTriangulation # for convex hull and triangulation
 import ExactPredicates
--- a/src/methods/convex_hull.jl
+++ b/src/methods/convex_hull.jl
@@ -1,0 +1,92 @@
+#=
+# Convex hull
+
+The _convex hull_ of a set of points is the smallest **convex** polygon that contains all the points.
+
+GeometryOps.jl provides a number of methods for computing the convex hull of a set of points, usually
+linked to other Julia packages.  
+
+## Example
+
+### Simple hull
+```@example simple
+import GeometryOps as GO, GeoInterface as GI
+using CairoMakie # to plot
+
+points = randn(GO.Point2f, 100)
+f, a, p = plot(points; label = "Points")
+hull_poly = GO.convex_hull(points)
+lines!(a, hull_poly; label = "Convex hull", color = Makie.wong_colors()[2])
+axislegend(a)
+f
+```
+
+## Convex hull of the USA
+```@example usa
+import GeometryOps as GO, GeoInterface as GI
+using CairoMakie # to plot
+using NaturalEarth # for data
+
+all_adm0 = naturalearth("admin_0_countries", 110)
+usa = all_adm0.geometry[findfirst(==("USA"), all_adm0.ADM0_A3)]
+f, a, p = lines(usa)
+lines!(a, GO.convex_hull(usa); color = Makie.wong_colors()[2])
+f
+```
+
+=#
+
+"""
+    convex_hull([method], geometries)
+
+Compute the convex hull of the points in `geometries`.  
+Returns a `GI.Polygon` representing the convex hull.
+
+Note that all 
+
+!!! warning
+    This interface only computes the 2-dimensional convex hull!
+
+    For higher dimensional hulls, use the relevant package (Qhull.jl or similar).
+"""
+function convex_hull end
+
+"""
+    DelaunayTriangulationMethod()
+
+This is an algorithm for the [`convex_hull`](@ref) function.
+
+Uses [`Delaunator.jl`](https://github.com/JuliaGeometry/DelaunayTriangulation.jl) to compute the convex hull.
+This is a pure Julia algorithm which provides an optimal Delaunay triangulation.
+
+See also [`convex_hull`](@ref)
+"""
+struct DelaunayTriangulationMethod end
+
+# , [`DelaunayTriangulationMethod`](@ref), [`EarCutMethod`](@ref).
+# GrahamScanMethod, etc. can be implemented in GO as well, if someone wants to...
+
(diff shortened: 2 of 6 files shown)


src/methods/convex_hull.jl
```diff
@@ -0,0 +1,92 @@
...
+    For higher dimensional hulls, use the relevant package (Qhull.jl or similar).
+"""
+function convex_hull end
+
+"""
+    DelaunayTriangulationMethod()
+
+This is an algorithm for the [`convex_hull`](@ref) function.
+
+Uses [`Delaunator.jl`](https://github.com/JuliaGeometry/DelaunayTriangulation.jl) to compute the convex hull.
```
DanielVandH, line 59: DelaunayTriangulation.jl

src/methods/convex_hull.jl
```diff
@@ -0,0 +1,92 @@
...
+struct DelaunayTriangulationMethod end
+
+# , [`DelaunayTriangulationMethod`](@ref), [`EarCutMethod`](@ref).
+# GrahamScanMethod, etc. can be implemented in GO as well, if someone wants to...
+
+convex_hull(geometries) = convex_hull(DelaunayTriangulationMethod(), geometries)
+
+# TODO: have this respect the CRS by pulling it out of `geometries`.
+function convex_hull(::DelaunayTriangulationMethod, geometries)
+    # Extract all points as tuples.  We have to collect and allocate
```
DanielVandH, line 73: Is this true? You can provide whatever point types you like so long as you extend the interfaces (which are hopefully sufficiently documented) from DelaunayTriangulation.jl appropriately. I give a complete example here https://juliageometry.github.io/DelaunayTriangulation.jl/dev/tutorials/custom_primitive/, maybe something goes wrong in `convex_hull` that I'm not aware of?

DanielVandH (commented):
Know it's in draft but happened to notice a couple of things here

src/methods/convex_hull.jl
asinghvi17: We have to flatten anyway to get a vector, and getting x and y values from C pointers is pretty bad for performance - best to do it once.  I can update the comment with better verbiage, though.

asinghvi17: Plus, we'd have to extend the interface for every GeoInterface compatible type, which would probably mean either piracy or changes on the DelaunayTriangulation end.  (Though if you do want to allow GeoInterface as a fallback it would be pretty easy I think).

DanielVandH: > We have to flatten anyway to get a vector, and getting x and y values from C pointers is pretty bad for performance - best to do it once. I can update the comment with better verbiage, though.

I have no preference either way just wanted to see if you were aware (I wasn't aware of the need for getting the values from C pointers, so fair enough). The verbiage is alright

> Plus, we'd have to extend the interface for every GeoInterface compatible type, which would probably mean either piracy or changes on the DelaunayTriangulation end. (Though if you do want to allow GeoInterface as a fallback it would be pretty easy I think).

It would probably be welcome as a package extension in DelaunayTriangulation. I'm not familiar with the interfaces though, so I probably won't be delving into that. Just leaving the code as is is fine anyway, it works and would be fast enough :).

asinghvi17: It would probably just be:
```julia
DT.getx(p) = DT.getx(GI.trait(p), p)
DT.gety(p) = DT.gety(GI.trait(p), p)
DT.number_type(p) = DT.number_type(GI.trait(p), p)

DT.getx(::GI.PointTrait, p) = GI.x(p)
DT.gety(::GI.PointTrait, p) = GI.y(p)
DT.number_type(::GI.PointTrait, p) = typeof(GI.x(p))

File: src/methods/convex_hull.jl (hunk 1 above)
Comment (asinghvi17) [reply src/methods/convex_hull.jl:59]:
```suggestion
Uses [`DelaunayTriangulation.jl`](https://github.com/JuliaGeometry/DelaunayTriangulation.jl) to compute the convex hull.

File: src/methods/convex_hull.jl [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+    DelaunayTriangulationMethod()
+
+This is an algorithm for the [`convex_hull`](@ref) function.
+
+Uses [`DelaunayTriangulation.jl`](https://github.com/JuliaGeometry/DelaunayTriangulation.jl) to compute the convex hull.
+This is a pure Julia algorithm which provides an optimal Delaunay triangulation.
+
+See also [`convex_hull`](@ref)
+"""
+struct DelaunayTriangulationMethod end
```
DanielVandH, line 64:
```suggestion
struct MonotoneChainMethod end
```

`DelaunayTriangulationMethod` sounds like you're doing `get_convex_hull(triangulate(points))` (i.e. getting the boundary of the unconstrained Delaunay triangulation) when really you're just doing the monotone chain algorithm which happens to be in DelaunayTriangulation.jl

src/methods/convex_hull.jl
```diff
@@ -0,0 +1,92 @@
...
+
+# , [`DelaunayTriangulationMethod`](@ref), [`EarCutMethod`](@ref).
```
DanielVandH, line 66:
```suggestion
# , [`MonotoneChainMethod`](@ref), [`EarCutMethod`](@ref).
```

src/methods/convex_hull.jl
```diff
@@ -0,0 +1,92 @@
...
+# GrahamScanMethod, etc. can be implemented in GO as well, if someone wants to...
+
+convex_hull(geometries) = convex_hull(DelaunayTriangulationMethod(), geometries)
```
DanielVandH, line 69:
```suggestion
convex_hull(geometries) = convex_hull(MonotoneChainMethod(), geometries)
```

src/methods/convex_hull.jl
```diff
@@ -0,0 +1,92 @@
...
+
+# TODO: have this respect the CRS by pulling it out of `geometries`.
+function convex_hull(::DelaunayTriangulationMethod, geometries)
```
DanielVandH, line 72:
```suggestion
function convex_hull(::MonotoneChainMethod, geometries)
```

src/GeometryOps.jl
```diff
@@ -4,16 +4,17 @@ module GeometryOps
 
 using GeoInterface
 using GeometryBasics
-import Tables
```
asinghvi17, line 7:
```suggestion
import Tables

**Comment by asinghvi17** [comment]
This is good to go from my end of things!

File: src/methods/convex_hull.jl [hunk 8]
```diff
@@ -0,0 +1,90 @@
...
+
+    # TODO: create a specialized method that extracts only the information
+    # required, GeometryBasics points can be passed through directly.
+    points = collect(flatten(tuples, GI.PointTrait, geometries))
+    # Compute the convex hull using DelTri (shorthand for DelaunayTriangulation.jl).
+    hull = DelTri.convex_hull(points)
+    # Convert the result to a `GI.Polygon` and return it.
+    # View would be more efficient here, but re-allocating
+    # is cleaner.
+    return GI.Polygon([GI.LinearRing(DelTri.get_points(hull)[(DelTri.get_vertices(hull))])])
```
rafaqz, line 89: Would be good to break this line up into two steps - get the points, make the polygon.

src/GeometryOps.jl
```diff
@@ -4,16 +4,18 @@ module GeometryOps
...
+import DelaunayTriangulation # for convex hull and triangulation
 import ExactPredicates
 import Base.@kwdef
-
-using GeoInterface.Extents: Extents
+import GeoInterface.Extents: Extents
 
 const GI = GeoInterface
 const GB = GeometryBasics
+const DelTri = DelaunayTriangulation
```
rafaqz, line 18:
```suggestion
const DT = DelaunayTriangulation
```

More consistent with the other package abbreviations

src/methods/convex_hull.jl
asinghvi17:
```suggestion
    point_vec = DelTri.get_points(hull)[(DelTri.get_vertices(hull))]
    return GI.Polygon([GI.LinearRing(point_vec)])

File: src/GeometryOps.jl (hunk 9 above)
Comment (asinghvi17) [reply src/GeometryOps.jl:18]: Since we won't be using DelaunayTriangulation all that much, I might just drop it in favour of a simple `import` and explicit `DelaunayTriangulation.*`...

**Comment by asinghvi17** [comment]
GEOS's convex hull is wound clockwise?  Somehow?  Simple features dictates the opposite but I guess this is following ESRI...

File: src/methods/convex_hull.jl [hunk 10]
```diff
@@ -0,0 +1,130 @@
...
+
+    # TODO: create a specialized method that extracts only the information
+    # required, GeometryBasics points can be passed through directly.
+    points = collect(flatten(tuples, GI.PointTrait, geometries))
+    # Compute the convex hull using DelTri (shorthand for DelaunayTriangulation.jl).
+    hull = DelaunayTriangulation.convex_hull(points)
+    # Convert the result to a `GI.Polygon` and return it.
+    # View would be more efficient here, but re-allocating
+    # is cleaner.
+    return GI.Polygon([GI.LinearRing(DelaunayTriangulation.get_points(hull)[(DelaunayTriangulation.get_vertices(hull))])])
```
asinghvi17, line 129:
```suggestion
    point_vec = DelaunayTriangulation.get_points(hull)[DelaunayTriangulation.get_vertices(hull)]
    return GI.Polygon([GI.LinearRing(point_vec)])

PR #164, opened 2024-06-20
Intersection points

skygering:
Before, the `intersection_points` function just skipped degenerate intersection points. This might not be the most performant way to do this due to the `unique!` call, but at least it is more correct than the previous version. 

If approved, I can also do a patch version bump.


src/methods/clipping/intersection.jl
```diff
@@ -175,57 +175,53 @@ function _intersection(
...
-                    push!(result, GI.Point(point))
-                end
-            end
+    # Loop over pairs of edges and add any unique intersection points to results
+    for a_edge in edges_a, b_edge in edges_b
+        line_orient, intr1, intr2 = _intersection_point(T, a_edge, b_edge; exact)
+        line_orient == line_out && continue
+        pt1, _ = intr1
+        push!(result, pt1)
+        if line_orient == line_over
```
rafaqz, line 215: We don't need to check line_cross or line_hinge any more?

asinghvi17 (approved):

PR #165, opened 2024-06-21
Describe how to navigate docs

asinghvi17:
I'm not sure what the best way to describe this is, but figured we should at least have something.


asinghvi17:
@thecedarprince this is what I was talking about yesterday.  Will go with this for now but hopefully we think of some better description/division soon.

PR #178, opened 2024-07-14
add DimensionalData extension

rafaqz:
This adds a DD extension for `polygonize` so that any DD array can be polygonized with X/Y (or custom with `dims` keyword) lookup values.

### DIFF
--- a/ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
+++ b/ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
@@ -11,7 +11,7 @@ function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs
     else
         @warn "`polygonsize` is not possible for `Points` sampling, as polygons cover space by definition. Treating as `Intervals`, but this may not be appropriate"
         map(lookups) do l
-            Dd.intervalbounds(DD.set(l, DD.Intervals()))
+            DD.intervalbounds(DD.set(l, DD.Intervals()))
         end
     end
     GO.polygonize(bounds_vecs..., DD.AbstractDimArray; crs, kw...)
--- a/ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
+++ b/ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
@@ -1,0 +1,21 @@
+module GeometryOpsDimensionalDataExt
+
+import DimensionalData as DD
+import GeometryOps as GO
+import GeoInterface as GI
+
+function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs(A), kw...)
+    lookups = DD.lookup(A, dims)
+    bounds_vecs = if DD.isintervals(lookups)
+        map(DD.intervalbounds, lookups)
+    else
+        @warn "`polygonsize` is not possible for `Points` sampling, as polygons cover space by definition. Treating as `Intervals`, but this may not be appropriate"
+        map(lookups) do l
+            Dd.intervalbounds(DD.set(l, DD.Intervals()))
+        end
+    end
+    GO.polygonize(bounds_vecs..., DD.AbstractDimArray; crs, kw...)
+end
+
+
+end
--- a/test/methods/polygonize.jl
+++ b/test/methods/polygonize.jl
@@ -1,5 +1,6 @@
 using Test
 import GeometryOps as GO
+import GeometryOps: polygonize
 import GeoInterface as GI
 import DimensionalData as DD
 import OffsetArrays, Rasters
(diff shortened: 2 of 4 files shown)


ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
```diff
@@ -0,0 +1,21 @@
...
+import GeoInterface as GI
+
+function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs(A), kw...)
+    lookups = DD.lookup(A, dims)
+    bounds_vecs = if DD.isintervals(lookups)
+        map(DD.intervalbounds, lookups)
+    else
+        @warn "`polygonsize` is not possible for `Points` sampling, as polygons cover space by definition. Treating as `Intervals`, but this may not be appropriate"
+        map(lookups) do l
+            Dd.intervalbounds(DD.set(l, DD.Intervals()))
```
asinghvi17, line 14: DD

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
```diff
@@ -0,0 +1,21 @@
...
+import GeometryOps as GO
+import GeoInterface as GI
+
+function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs(A), kw...)
+    lookups = DD.lookup(A, dims)
+    bounds_vecs = if DD.isintervals(lookups)
+        map(DD.intervalbounds, lookups)
+    else
+        @warn "`polygonsize` is not possible for `Points` sampling, as polygons cover space by definition. Treating as `Intervals`, but this may not be appropriate"
+        map(lookups) do l
```
asinghvi17, line 13: Is this mutating the input array?  Do we want to copy instead?

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
```diff
@@ -0,0 +1,21 @@
+module GeometryOpsDimensionalDataExt
+
+import DimensionalData as DD
+import GeometryOps as GO
+import GeoInterface as GI
+
+function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs(A), kw...)
```
asinghvi17, line 7: What happens with the dims that aren't given (time etc)?

asinghvi17:
Does that last commit set minimum version to Julia 1.10 instead of 1.9?

rafaqz:
Oops I used the wrong commit message. Can fix that in the squash later

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
rafaqz: No bang! no mutation

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
rafaqz: Yeah I haven't done that yet, this should be WIP.

But I image we make a feature collection with a feature for each slice, and put the lookup values as properties?

rafaqz:
Oops wrong commit message, we can fix it in the squash

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
asinghvi17:
```suggestion
            DD.intervalbounds(DD.set(l, DD.Intervals()))
```

test/methods/polygonize.jl
```diff
@@ -51,19 +55,23 @@ end
...
-        evil_mp = @test_nowarn polygonize(evil)
-        @test GO.equals(data_mp, evil_mp)
+        data_mp = polygonize(51:150, 151:200, data);
+        dd_mp = polygonize(dd);
+        @test GO.equals(data_mp, dd_mp)
     end
+
     @testset "Rasters" begin
-        data = rand(1:4, 100, 100) .== 1
-        evil = Rasters.Raster(data; dims = (DimensionalData.X(1:100), DimensionalData.Y(1:100)), crs = Rasters.GeoFormatTypes.EPSG(4326))
```
asinghvi17, line 63: Oh no, the evil is gone :D

rafaqz: Turned out it wasn't actually evil, just a regular garden variety array

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
rafaqz: Actually this is a case for R style vector data cubes.

But we can just broadcast over the other dimensions and get a DimArray of X/Y geometries. 

We should be able to write that directly as a FeatureCollection with GeoJSON/ArchGDAL/Shapefile without any more work as long as we name the array `:geometry`.

asinghvi17: Do we then want to add an assert here that the dims can only have length 2?

asinghvi17:
Is this just waiting on the datacube stuff/implementation?

rafaqz:
Waiting for me to have time 😭

ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
```diff
@@ -0,0 +1,21 @@
...
+    lookups = DD.lookup(A, dims)
+    bounds_vecs = if DD.isintervals(lookups)
+        map(DD.intervalbounds, lookups)
+    else
+        @warn "`polygonsize` is not possible for `Points` sampling, as polygons cover space by definition. Treating as `Intervals`, but this may not be appropriate"
```
asinghvi17, line 12: Doesn't this just mean that the ring vertices are located at the points?  That seems pretty exact...

rafaqz: Hmm not sure I understand you. It's just saying that points don't cover any space so polygonizing them doesn't make sense

asinghvi17: I was thinking of it more in a sense that the polygonizing operation is the concave hull of the set of points that satisfy `f`...

rafaqz: Yeah, somehing else! `polygonize` is like a groupby/combine for lots of little square polygons

PR #180, opened 2024-07-19
fix typos

spaette:
Is the intention `geometries'` or `geometry's` here?

```
$ ed -s GeometryOps.jl/src/methods/geom_relations/disjoint.jl <<<'173,178p'
#= Geometry is disjoint from a linestring if the line's interior and boundary
points don't intersect with the geometrie's interior and boundary points. =#
_disjoint(
    trait1::Union{GI.LinearRingTrait, GI.PolygonTrait}, g1,
    trait2::Union{GI.LineTrait, GI.LineStringTrait}, g2,
) = _disjoint(trait2, g2, trait1, g1)
$ 
```

The pull has the typo fixed but the wording is a bit awkward here.

```
$ ed -s GeometryOps.jl/src/methods/area.jl <<<'79p'
This is computed slighly differently for different geometries:
$ 
```


asinghvi17 (approved):
Thanks a lot, @spaette !  Hope you didn't have to do all of this by hand :)

asinghvi17:
The intention there was for the singular form so that works.

asinghvi17:
The wording for area should probably read "computed differently for different geometry types"

spaette:
> by hand

The work was based on dumped repo strings.

> "computed differently for different geometry types"

How about rewording as follows?

`For different geometries the computation is slightly modified:`

spaette:
@asinghvi17 

Go ahead and merge.

src/methods/area.jl

I'm leaving any debate on a possible rewrite to the Maintainers.

asinghvi17:
Hmm, that makes sense.  I am refactoring a bunch of stuff anyway so will sneak a rewrite in there.

spaette:
if the line is changed 

one might want a potential rewrite to be suitable to both circumstances where the current line occurs 

```
$ ed -s GeometryOps.jl/src/methods/area.jl <<<'53,68p'
"""
    area(geom, [T = Float64])::T

Returns the area of a geometry or collection of geometries. 
This is computed slightly differently for different geometries:

    - The area of a point/multipoint is always zero.
    - The area of a curve/multicurve is always zero.
    - The area of a polygon is the absolute value of the signed area.
    - The area multi-polygon is the sum of the areas of all of the sub-polygons.
    - The area of a geometry collection, feature collection of array/iterable 
        is the sum of the areas of all of the sub-geometries. 

Result will be of type T, where T is an optional argument with a default value
of Float64.
"""
$ ed -s GeometryOps.jl/src/methods/area.jl <<<'75,91p'
"""
    signed_area(geom, [T = Float64])::T

Returns the signed area of a single geometry, based on winding order. 
This is computed slightly differently for different geometries:

    - The signed area of a point is always zero.
    - The signed area of a curve is always zero.
    - The signed area of a polygon is computed with the shoelace formula and is
    positive if the polygon coordinates wind clockwise and negative if
    counterclockwise.
    - You cannot compute the signed area of a multipolygon as it doesn't have a
    meaning as each sub-polygon could have a different winding order.

Result will be of type T, where T is an optional argument with a default value
of Float64.
"""
$ 
```

PR #182, opened 2024-07-21
[WIP] Spherical triangulation and interpolation

asinghvi17:
This works, but the code is a bit messy and I implemented a bunch of things myself.  Would love to return a custom point type for spherical coordinates that encodes Cartesian coords on the unit sphere, but that would be a bit much for an experiment.  Maybe when we actually do it :D 

https://github.com/user-attachments/assets/1d14e937-54db-45c0-b362-b33f912ae75e


docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,168 @@
...
+triangulation.points[end-2:end] .= (stereographic_points[pivot_ind],)
+
+f, a, p = triplot(triangulation; axis = (; type = Axis3,));
+m = p.plots[1].plots[1][1][]
+
+geo_mesh_points = vcat(points, repeat([pivot_point], 3))
+cartesian_mesh_points = UnitCartesianFromGeographic().(geo_mesh_points)
+f, a, p = mesh(cartesian_mesh_points, getfield(getfield(m, :simplices), :faces))
+wireframe(p.mesh[])
+CairoMakie.GeometryBasics.Mesh(UnitCartesianFromGeographic().(vcat(points, repeat([pivot_point], 3))), splat(CairoMakie.GeometryBasics.TriangleFace).(collect(triangulation.triangles))) |> wireframe
```
DanielVandH, line 62: You can also avoid the ghost triangles here by doing `each_solid_triangle(triangulation)` instead (`.triangles` isn't the public API for accessing the triangles).

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,168 @@
...
+# Add infinite horizon points
+triangulation_points = copy(stereographic_points)
+push!(triangulation_points, Point2(0, far_value))
+push!(triangulation_points, Point2(-far_value, 0))
+push!(triangulation_points, Point2(0, -far_value))
+
+triangulation = DelTri.triangulate(triangulation_points)
+
+triangulation.points[1:end-3] .= stereographic_points
+triangulation.points[end-2:end] .= (stereographic_points[pivot_ind],)
```
DanielVandH, line 53: If you want to stick the API, you could also here use `get_points(triangulation)` and work with that result instead of using `.points` which I don't think I've put as the API (not intentionally anyway).

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,168 @@
...
+f, a, p = triplot(triangulation; axis = (; type = Axis3,));
+m = p.plots[1].plots[1][1][]
+
+geo_mesh_points = vcat(points, repeat([pivot_point], 3))
+cartesian_mesh_points = UnitCartesianFromGeographic().(geo_mesh_points)
+f, a, p = mesh(cartesian_mesh_points, getfield(getfield(m, :simplices), :faces))
+wireframe(p.mesh[])
+CairoMakie.GeometryBasics.Mesh(UnitCartesianFromGeographic().(vcat(points, repeat([pivot_point], 3))), splat(CairoMakie.GeometryBasics.TriangleFace).(collect(triangulation.triangles))) |> wireframe
+
+geo_triangulation = deepcopy(triangulation)
```
DanielVandH, line 64: I don't think it'll cause any problems for your application but I think `deepcopy` on `Triangulation`s might not be currently working 100% correctly due to some aliasing between fields. I have an issue for this here https://github.com/JuliaGeometry/DelaunayTriangulation.jl/issues/142. Is it 100% needed?

If it does cause issues you can also use `retriangulate(triangulation, geo_mesh_points)`, though this might be a bit slow

DanielVandH (commented):
Just some comments about API. Nothing too important since this is just an experiment anyway

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: It's not really needed, but this was the quick and dirty solution without having to look at the docs :D.

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: I'll probably plot the triangles with faces from the triangulation and precomputed coordinates, this was more of an experiment (which failed, unfortunately)

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: This is exactly what I was missing - thanks!

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,332 @@
...
+    # triangulation = DelTri.triangulate(triangulation_points; boundary_nodes)
+    triangulation = DelTri.triangulate(triangulation_points)
+    #  for diagnostics, try `fig, ax, sc = triplot(triangulation, show_constrained_edges=true, show_convex_hull=true)`
+
+    # First, get all the "solid" faces, ie, faces not attached to boundary nodes
+    boundary_faces = Iterators.filter(Base.Fix1(DelTri.is_boundary_triangle, triangulation), DelTri.each_solid_triangle(triangulation)) |> collect
+    faces = map(facetype, DelTri.each_solid_triangle(triangulation))
+
+    for ghost_face in DelTri.each_ghost_triangle(triangulation)
+        push!(faces, facetype(map(ghost_face) do i; i ≤ 0 ? pivot_ind : i end))
```
DanielVandH, line 68:
```suggestion
        push!(faces, facetype(map(ghost_face) do i; DelaunayTriangulation.is_ghost_vertex(i) ? pivot_ind : i end))
```

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,332 @@
...
+lonlat2bertin = Proj.Transformation(GFT.EPSG(4326), GFT.ProjString("+proj=bertin1953 +type=crs +lon_0=-16.5 +lat_0=-42 +x_0=7.93 +y_0=0.09"); always_xy = true)
+lons = LinRange(-180, 180, 300)
+lats = LinRange(-90, 90, 150)
+bertin_points = lonlat2bertin.(lons, lats')
+
+projected_points = GO.reproject(GO.tuples(points), source_crs = GFT.EPSG(4326), target_crs = "+proj=bertin1953 +lon_0=-16.5 +lat_0=-42 +x_0=7.93 +y_0=0.09")
+
+ch = DelTri.convex_hull(projected_points) # assumes each point is in the triangulation
+boundary_nodes = DelTri.get_vertices(ch) 
+bertin_boundary_poly = GI.Polygon([GI.LineString(DelTri.get_points(ch)[DelTri.get_vertices(ch)])])
```
DanielVandH, line 96: What do you need `bertin_boundary_poly` for? Are you just using it to see if a point is inside the triangulation? You could look at `DelaunayTriangulation.dist(tri, q)` I think

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,332 @@
...
+        itp(x, y; method = Nearest())
+    else
+        NaN
+    end
+    for (x, y) in bertin_points
+]
+# TODO: this currently doesn't work, because some points are not inside a triangle and so cannot be located.
+# Options are:
+# 1. Reject all points outside the convex hull of the projected points.  (Tried but failed)
+# 2. ...
```
DanielVandH, line 112: What goes wrong here? Can you just do something like

```julia
xv, yv = first.(bertin_points), last.(bertin_points)
vals = itp(xv, yv, method = Nearest(), project = false) # Nearest() doesn't seem to use project...
vals[identify_exterior_points(xv, yv, itp)] .= NaN 
```

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,332 @@
...
+for (i, line) in enumerate(stereographic_grid)
+    lines!(ax, line; linewidth = 3, color = Makie.resample_cmap(:viridis, length(stereographic_grid))[i])
+end
+fig
+
+all(reduce(vcat, faces) |> sort! |> unique! .== 1:length(points))
+
+faces[findall(x -> 1 in x, faces)]
+
+# try NaturalNeighbours, fail miserably
```
DanielVandH, line 213: Are there issues even when using `Sibson(1)` below? Or what's the misery

DanielVandH (commented):
Just a quick look for now. Did you have any luck trying to use ghost vertices?

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17:
```
ERROR: TaskFailedException

    nested task error: The point, (-9.986746533293853e6, -6.557879586899938e6), could not be located.
    Stacktrace:
      [1] brute_force_search(tri::DelaunayTriangulation.Triangulation{…}, q::Tuple{…}; itr::Set{…})
        @ DelaunayTriangulation ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/brute_force.jl:36
      [2] brute_force_search
        @ ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/brute_force.jl:31 [inlined]
      [3] restart_jump_and_march(tri::DelaunayTriangulation.Triangulation{…}, q::Tuple{…}, store_history::Val{…}, history::Nothing, rng::Random.TaskLocalRNG, maxiters::Int64, cur_iter::Int64, concavity_protection::Bool, num_restarts::Int64, use_barriers::Val{…})
        @ DelaunayTriangulation ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/jump_and_march.jl:1236
      [4] _jump_and_march(tri::DelaunayTriangulation.Triangulation{…}, q::Tuple{…}, k::Int64, store_history::Val{…}, history::Nothing, rng::Random.TaskLocalRNG, maxiters::Int64, cur_iter::Int64, concavity_protection::Bool, num_restarts::Int64, use_barriers::Val{…})
        @ DelaunayTriangulation ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/jump_and_march.jl:1145
      [5] restart_jump_and_march(tri::DelaunayTriangulation.Triangulation{…}, q::Tuple{…}, store_history::Val{…}, history::Nothing, rng::Random.TaskLocalRNG, maxiters::Int64, cur_iter::Int64, concavity_protection::Bool, num_restarts::Int64, use_barriers::Val{…})
        @ DelaunayTriangulation ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/jump_and_march.jl:1234
    --- the last 2 lines are repeated 23 more times ---
     [52] _jump_and_march(tri::DelaunayTriangulation.Triangulation{…}, q::Tuple{…}, k::Int64, store_history::Val{…}, history::Nothing, rng::Random.TaskLocalRNG, maxiters::Int64, cur_iter::Int64, concavity_protection::Bool, num_restarts::Int64, use_barriers::Val{…})
        @ DelaunayTriangulation ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/jump_and_march.jl:1145
     [53] #jump_and_march#87
        @ ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/jump_and_march.jl:739 [inlined]
     [54] jump_to_voronoi_polygon(tri::DelaunayTriangulation.Triangulation{…}, q::Tuple{…}; kwargs::@Kwargs{…})
        @ DelaunayTriangulation ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/nearest_neighbour.jl:24
     [55] jump_to_voronoi_polygon
        @ ~/.julia/dev/DelaunayTriangulation/src/algorithms/point_location/nearest_neighbour.jl:23 [inlined]
     [56] _compute_nearest_coordinates(tri::DelaunayTriangulation.Triangulation{…}, interpolation_point::Tuple{…}, cache::NaturalNeighbours.NaturalNeighboursCache{…}; kwargs::@Kwargs{…})
        @ NaturalNeighbours ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/coordinates/nearest.jl:10
     [57] _compute_nearest_coordinates
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/coordinates/nearest.jl:1 [inlined]
     [58] #compute_natural_coordinates#28
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/coordinates/nearest.jl:19 [inlined]
     [59] compute_natural_coordinates
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/coordinates/nearest.jl:18 [inlined]
     [60] #_eval_interp#21
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/eval.jl:42 [inlined]
     [61] _eval_interp
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/eval.jl:40 [inlined]
     [62] (::NaturalNeighbours.NaturalNeighboursInterpolant{…})(x::Float64, y::Float64, id::Int64; parallel::Bool, method::Nearest{…}, kwargs::@Kwargs{…})
        @ NaturalNeighbours ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/interpolate.jl:187
     [63] NaturalNeighboursInterpolant
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/interpolate.jl:180 [inlined]
     [64] macro expansion
        @ ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/interpolate.jl:204 [inlined]
     [65] (::NaturalNeighbours.var"#71#threadsfor_fun#18"{…})(tid::Int64; onethread::Bool)
        @ NaturalNeighbours ./threadingconstructs.jl:215
     [66] #71#threadsfor_fun
        @ ./threadingconstructs.jl:182 [inlined]

...and 7 more exceptions.

Stacktrace:
 [1] threading_run(fun::NaturalNeighbours.var"#71#threadsfor_fun#18"{…}, static::Bool)
   @ Base.Threads ./threadingconstructs.jl:172
 [2] macro expansion
   @ ./threadingconstructs.jl:220 [inlined]
 [3] (::NaturalNeighbours.NaturalNeighboursInterpolant{…})(vals::Vector{…}, x::Vector{…}, y::Vector{…}; parallel::Bool, method::Nearest{…}, kwargs::@Kwargs{…})
   @ NaturalNeighbours ~/.julia/packages/NaturalNeighbours/HJ2P0/src/interpolation/interpolate.jl:202
```

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: I was trying to see if I could filter out points that are outside the convex hull and therefore not within any triangle of `tri`, but no luck with that.

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: Yeah I'm not quite sure what the issue is here, in the worst case I can always sample in stereographic space but that seems pretty bad...

docs/src/experiments/spherical_delaunay_stereographic.jl
DanielVandH: That's... concerning. That should be impossible since it should be finding it in a ghost triangle at the very least... will take a look and actually run the code later.

asinghvi17: I think one problem is that the mesh and convex hull here don't actually cover the full Earth - see `mesh(bertin_points, faces)`
download-12

asinghvi17: see https://proj.org/en/9.4/operations/projections/bertin1953.html or https://observablehq.com/@fil/spherical-contours?collection=@fil/voronoi for comparison

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,332 @@
...
+        triangulation_points[really_far_idxs] .= (Point2(far_value, 0.0),)
+    end
+
+    boundary_points = reverse([
+        (-far_value, -far_value),
+        (-far_value, far_value),
+        (far_value, far_value),
+        (far_value, -far_value),
+        (-far_value, -far_value),
+    ])
```
DanielVandH, line 55: This is unused. Were there issues? You mentioned an issue with `convert_boundary_points_to_indices`, is it related?

DanielVandH: `boundary_faces` is also unused

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,332 @@
...
+# with the same faces, but in a Bertin projection...
+using NaturalNeighbours
+lonlat2bertin = Proj.Transformation(GFT.EPSG(4326), GFT.ProjString("+proj=bertin1953 +type=crs +lon_0=-16.5 +lat_0=-42 +x_0=7.93 +y_0=0.09"); always_xy = true)
+lons = LinRange(-180, 180, 300)
+lats = LinRange(-90, 90, 150)
+bertin_points = lonlat2bertin.(lons, lats')
+
+projected_points = GO.reproject(GO.tuples(points), source_crs = GFT.EPSG(4326), target_crs = "+proj=bertin1953 +lon_0=-16.5 +lat_0=-42 +x_0=7.93 +y_0=0.09")
+
+ch = DelTri.convex_hull(projected_points) # assumes each point is in the triangulation
```
DanielVandH, line 94: @asinghvi17 `projected_points` is three-dimensional. The `convex_hull` function is treating it by ignoring the `z` coordinate of each point. What is supposed to be the input here? This is the issue you're having with NaturalNeighbours - the triangulation is very invalid.

image

asinghvi17: `projected_points`' z coordinate only carries metadata, so it's fine for it to be three-dimensional.  That being said I haven't checked the triangulation in stereographic space, it might have issues there too.

asinghvi17: On main the boundary points generate degenerate triangles, I guess I can remove those?

DanielVandH: I'm surprised that it's generating triangles where some of the vertices repeat..

asinghvi17: That's not entirely unexpected, since the triangulation is computed in stereographic space (where the triangulation generated _is_ planar, correct, etc) and not in the projected space that we are currently in.

DanielVandH: I see. I'm not familiar enough with stereographic projections, I thought they would preserve the structure. I don't see a good way to work with the 2D triangulation in that case, since it isn't actually a triangulation.

docs/src/experiments/spherical_delaunay_stereographic.jl
```diff
@@ -0,0 +1,351 @@
+#=
```
DanielVandH, line 1: Probably also need to include 

```
Copyright 2018-2021 Philippe Rivière

Permission to use, copy, modify, and/or distribute this software for any purpose
with or without fee is hereby granted, provided that the above copyright notice
and this permission notice appear in all copies.

THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
THIS SOFTWARE.
```

here from https://github.com/Fil/d3-geo-voronoi/blob/main/LICENSE if this code is based on it

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: Yes, that would be a problem.  Maybe it's best to create some kind of spherical triangulation type, which can also handle spherical Voronoi?  That sounds like a lot of effort though.

DanielVandH: I don't really know how you would get around this issue. It just seems like a bug somewhere to me.. the projection isn't an isometry but it should be conformal, right? I think conformal maps would show up better than what we see in the image above, e.g. it should preserve angles. The crossings seem to come only from boundary vertices (from a visual check anyway, who knows about the interior) which suggests you're missing something when handling only the interior triangles.

asinghvi17: Ah, to clarify - the triangulation is done in stereographic space, where the triangulation would be completely valid.  I'm trying to interpolate in the Bertin 1953 projection, which doesn't have the same conformal property as the stereographic projection.  I guess what I was trying to do is actually the incorrect thing, and I should be figuring out how to interpolate on the sphere instead...

docs/src/experiments/spherical_delaunay_stereographic.jl
asinghvi17: The issue here was that a lot of faces were connecting to two boundary points and one data point, which is effectively a degenerate face and thus useless.

docs/src/experiments/spherical_delaunay_stereographic.jl
DanielVandH: I see.. my thinking was that you can do everything on the valid triangulation and then lift it back up to the sphere. I'll probably have to look more at the algorithm itself to understand.

asinghvi17: At least for interpolation, I would expect that to have issues when getting the values of points near the chosen pivot point in spherical space - there's always a point which wouldn't be connected to the triangulation in stereographic space, and neighbours in spherical space would be separated by almost the entire domain of points.

The first problem could be solved by injecting a rectangular boundary, but I'm not really sure how to deal with the second problem.

asinghvi17:
@DanielVandH is it possible to re-implement Laplace coordinates if I simply insert points into the triangulation in stereographic space?  Consider the following workflow:
- Triangulate in stereographic space and retain stereographic points
- Take interpolation points as (long, lat) and transform back to stereographic space, then compute Bowyer envelope
- Now that I have envelope vertex data, run a Laplace interpolation on the sphere.  This seems relatively straightforward, and we can avoid all of the issues with computing the actual Voronoi tessellation and area stealing etc. inherent to "pure" Sibson coordinates.

DanielVandH:
That would probably work. Might not be truly natural neighbour interpolation from the sphere (I think? I don't know its definition on the sphere exactly) but it would be good enough I reckon. Let me know if it's not clear how to use some of the functions in NaturalNeighbours.jl for this. I can also make `public` whatever internals you might need.

asinghvi17:
Yeah, that's what I was afraid of :(.  There might be a way to mimic the part of the Bowyer-Watson algorithm we need on the sphere, so I might be able to reimplement natural neighbour implementation on the sphere with a few minor adjustments.

Will give implementing that a try and report back shortly!

DanielVandH:
The envelope would be the same, i.e. the natural neighbours from the envelope in 2D _are_ actually the natural neighbours in 2D. It's the actual computation of the Laplacian coordinates that could differ. So maybe this is in some way a natural neighbour interpolant.

Remember that the Laplace coordinates are defined by 
```math 
\lambda_i = \frac{\hat\lambda_i}{\sum_j \hat\lambda_j},
```
where the sum of over the set of all natural neighbours and 
```math
\hat\lambda_i = \frac{\ell(\mathcal F_{0i})}{||\boldsymbol x_i - \boldsymbol x_j||},
```
where $\ell(\mathcal F_{0i})$ is the length of the interface that the Voronoi cell from the inserted point shares with $\boldsymbol x_j$. See the image below.
new_tile
Maybe Laplacian interpolation on a sphere is just computing the length of the inverse of that blue interface under the stereographic projection, and $||boldsymbol x_i - \boldsymbol x_j||$ is interpreted similarly. I don't know if those details are difficult to get right or not, but just some thoughts.

asinghvi17:
That sounds right to me - spherical distance is just an inverse cosine of the dot product between Cartesian vectors, so relatively easy to compute.  Circumcenters and circumradii are also easy to find, and I might end up caching them since the number of interpolation points is not necessarily very large.  

I don't think that this proof will extend to insertions, because we have to modify the triangulation when going back to the sphere.  Specifically, since we rotate about a pivot point, we have to change the ghost vertex to point back to the pivot point.  Since the ghost vertex is at infinity, inserting a point on one side of the triangulation will not change the other side, even if it should.  I'm not 100% sure about this and will have to do some testing, but I suspect it is the case.


I got _something_ with the spherical Laplace approach, though there are a lot of NaNs coming through:
download-5

asinghvi17:
Fixed one bug, this looks better now:
download-6

DanielVandH:
> I don't think that this proof will extend to insertions, because we have to modify the triangulation when going back to the sphere. Specifically, since we rotate about a pivot point, we have to change the ghost vertex to point back to the pivot point. Since the ghost vertex is at infinity, inserting a point on one side of the triangulation will not change the other side, even if it should. I'm not 100% sure about this and will have to do some testing, but I suspect it is the case.

Ah. You're probably right.. wasn't thinking about the infinity issues. It'd be interesting to see how natural neighbour interpolation on a sphere is actually done (I don't have access to that book you asked for on Slack either) to know what to do.

I do some NaN handling in some places in NaturalNeighbours.jl I think. Might want to search for `isnan` in the repo for some ideas

asinghvi17:
Just fixed the winding order, I still have some strange looking circumcircle artifacts but this is looking a lot more reasonable than before :D 
download

asinghvi17:
Here's a test case: 10 random points on the sphere (value 0), plus one at the north pole (value 1).

Screenshot 2024-07-31 at 7 18 47 AM

Will probably help quite a bit with debugging...

PR #190, opened 2024-07-30
Fix polygonize with LinRange

### LINKED ISSUE #187: Polygonize example is broken
Trying out the example here: https://juliageo.org/GeometryOps.jl/stable/source/methods/polygonize#Polygonizing-raster-data

Results in an error

```julia
julia> using GeometryOps, GLMakie, GeoMakie

julia> n = 49
49

julia> xs, ys = LinRange(-3, 3, n), LinRange(-3, 3, n);

julia> zs = Makie.peaks(n);

julia> polygons = polygonize(xs, ys, 0.8 .< zs .< 3.2)
ERROR: MethodError: no method matching similar(::Type{Vector{Tuple{Float64, Float64}}}, ::LinRange{Float64, Int64})
```

asinghvi17:
Fix #187

### DIFF
--- a/src/methods/polygonize.jl
+++ b/src/methods/polygonize.jl
@@ -121,8 +121,8 @@ function _polygonize(f::Base.Callable, xs::AbstractRange, ys::AbstractRange, A::
     ybounds = first(ys) - yhalf : step(ys) : last(ys) + yhalf
     Tx = eltype(xbounds)
     Ty = eltype(ybounds)
-    xvec = similar(Vector{Tuple{Tx,Tx}}, xs)
-    yvec = similar(Vector{Tuple{Ty,Ty}}, ys)
+    xvec = similar(Vector{Tuple{Tx,Tx}}, axes(xs))
+    yvec = similar(Vector{Tuple{Ty,Ty}}, axes(ys))
     for (xind, i) in enumerate(eachindex(xvec))
         xvec[i] = xbounds[xind], xbounds[xind+1]
     end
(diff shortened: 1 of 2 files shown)


rafaqz:
Did you load OffsetArrays.jl and see if that still works?

rafaqz:
I think we should get rid of `similar` and just define the undef vector

asinghvi17:
Yes, this seems to work with OffsetArrays loaded, but I'll add a test anyway

src/methods/polygonize.jl
```diff
@@ -121,8 +121,8 @@ function _polygonize(f::Base.Callable, xs::AbstractRange, ys::AbstractRange, A::
     ybounds = first(ys) - yhalf : step(ys) : last(ys) + yhalf
     Tx = eltype(xbounds)
     Ty = eltype(ybounds)
-    xvec = similar(Vector{Tuple{Tx,Tx}}, xs)
-    yvec = similar(Vector{Tuple{Ty,Ty}}, ys)
+    xvec = similar(Vector{Tuple{Tx,Tx}}, axes(xs))
```
rafaqz, line 124:
```suggestion
    xvec = Vector{Vector{Tuple{Tx,Tx}}}(undef, length(axes(xs)))
```

Wont something like this work and skip the complexity of `similar`?

tiemvanderdeure:
Now that https://github.com/JuliaGeo/GeometryOps.jl/pull/202 is merged we should close this

PR #202, opened 2024-09-04
fix error with undefined similar in polygonize

tiemvanderdeure:
polygonize calls `similar` on a Type and the abstractranges provided, but `similar` is not defined for all `AbstractRange` types. 

I think the easiest fix is just to call `length` on the range first.

### DIFF
--- a/src/methods/polygonize.jl
+++ b/src/methods/polygonize.jl
@@ -121,8 +121,8 @@ function _polygonize(f::Base.Callable, xs::AbstractRange, ys::AbstractRange, A::
     ybounds = first(ys) - yhalf : step(ys) : last(ys) + yhalf
     Tx = eltype(xbounds)
     Ty = eltype(ybounds)
-    xvec = similar(Vector{Tuple{Tx,Tx}}, xs)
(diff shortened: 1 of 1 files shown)


asinghvi17:
Looks good, thanks!  Do you have a test case that failed, so we can add it to the tests?

tiemvanderdeure:
I just noticed there is a very similar PR that never got merged https://github.com/JuliaGeo/GeometryOps.jl/pull/190

Maybe this is not as simple as I thought?

tiemvanderdeure:
I just went ahead and added tests, and let's see what happens.

I wanted to add a better test to make sure this returns the same polygon with different xs and ys like so:

```
    data_mp = polygonize(data)
    data_mp_range200 = polygonize(2:2:200, 2:2:200, data)
    data_mp_range200_2 = GO.apply(GI.PointTrait, data_mp_range200) do p
        p ./ 2
    end
    @test GO.equals(data_mp_range200_2, data_mp)
```

but this fails.

rafaqz (approved):

PR #206, opened 2024-09-17
[WIP] Tutorial to clip a feature collection

asinghvi17:


asinghvi17:
The coverage code would be nice to create a clipping recipe with, maybe using extents.  We can get away with losing a lot of exactness since it's not required (one of the polygons is guaranteed to have axis aligned edges, after all).

https://github.com/JuliaGeo/GeometryOps.jl/blob/main/src/methods/clipping/coverage.jl

PR #211, opened 2024-09-19
Add DataAPI metadata passthrough to `apply`

asinghvi17:
This PR propagates all non-`:default` style metadata through to the result table in `apply` on tables, if:
- there is metadata that can be read in the input table
- the result table type supports metadata writes (so not a named tuple table, for example)

This PR also adds `GEOINTERFACE:geometrycolumns` and `GEOINTERFACE:crs` tags to the resulting table, and ensures that the CRS is correct according to:
a. the `crs` kwarg to `apply`
if that is not present,
b. `GI.crs(input_table)`

this way, `GO.reproject` automatically changes the table metadata too!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.11"
 
 [deps]
 CoordinateTransformations = "150eb455-5306-5404-9cee-2592286d6298"
+DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 DelaunayTriangulation = "927a84f5-c5f4-47a5-9785-b46e178433df"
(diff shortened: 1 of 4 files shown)


asinghvi17:
I should probably PR this idea to GeoInterface too.

rafaqz (approved):
Looks good.

It could be nice to put this metadata update function in GeoInterface.jl some day, so we use the same thing everywhere. But here is fine for now.

asinghvi17:
Agreed, will look at a GeoInterface PR to encode this standard tomorrow.  Then we can go about defining metadata for e.g. GeoJSON, Shapefile, etc so everything just flows.

PR #220, opened 2024-10-02
Factor out primitives and types to `GeometryOpsCore.jl`

asinghvi17:
This PR creates a new GeometryOpsCore.jl package that other packages can depend on so we have a common core of types and frameworks throughout the ecosystem.

cc @rafaqz

This is still VERY WIP, we may choose to put more functionality in (I do want table support for flatten etc) or take functionality out.

Ideally, this goes out as a first draft in the next few days so it can be registered in the General registry.

This will enable the `cellarea(Spherical())` idea for https://github.com/rafaqz/Rasters.jl/pull/768.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /Manifest.toml
+/GeometryOpsCore/Manifest.toml
 /docs/build/
 /docs/src/source/
 .vscode/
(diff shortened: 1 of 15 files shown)


rafaqz (approved):
LGTM. Let's get the registration rolling we can fix the details later

asinghvi17:
```
┌ GeometryOps
│  WARNING: could not import Base.nthreads into GeometryOpsCore
│  WARNING: could not import Base.@threads into GeometryOpsCore
│  WARNING: could not import Base.@spawn into GeometryOpsCore
└  
```

is that just because of the double / triple import?

asinghvi17:
Will merge if docs build correctly now.

PR #223, opened 2024-10-06
[WIP] Use GeometryOpsCore for real

asinghvi17:
This PR switches GeometryOps to actually import the released version of GeometryOpsCore.

It turns out that this works on 1.10, but fails on nightly because it imports anonymous functions as well, and `gensym` doesn't know that so generates identical symbols for different anonymous functions in GeometryOps and GeometryOpsCore.  I fix that by creating a static import list of things that GeometryOpsCore defines, are unexported, but must be defined in GeometryOps.

Additionally, the PR will deprecate the GeodesicSegments and LinearSegments methods for `Geodesic` and `Linear` manifolds.  This is in preparation for the next breaking release where we'll introduce perimeter calculations, arclength interpolation, and basic spherical geometry utilities.


src/GeometryOps.jl
```diff
@@ -2,11 +2,19 @@
 
 module GeometryOps
 
-include("../GeometryOpsCore/src/GeometryOpsCore.jl") # TODO: replace this with `using GeometryOpsCore`
-import .GeometryOpsCore
-for name in setdiff(names(GeometryOpsCore, all = true), (:eval, :var"#eval", :include, :var"#include"))
+import GeometryOpsCore
+for name in Base.setdiff(
```
rafaqz, line 6: This is pretty obscure, is there a less scary way to do this so regular contributors don't run away straight away?

Personally I would just make a single tuple with everything you want to import and skip the union/setdiff part.

We should also export everything explicitly so people can read it

asinghvi17: Ah yeah that's a good point.  I can probably just explicitly import stuff.

tiemvanderdeure:
What is the status on this? I thought about getting started with adding some basic spherical geometry operations like distance and centroid, but maybe I should wait for these imports to be resolved?

asinghvi17:
I'm hoping to get this resolved next week feiw, pretty sure it's just a bad kwarg I have to track down

GeometryOpsCore/src/types.jl
```diff
@@ -51,6 +51,8 @@ end
 
 A spherical manifold means that the geometry is on the 3-sphere (but is represented by 2-D longitude and latitude).  
 
+By default, the radius is defined as the mean radius of the Earth, `6371008.8 m`.
```
rafaqz, line 54: Can we make global named constants for all the numbers in these files, and both insert them in dics and use them for keywords?

Maybe in a different PR, just commenting because this text is added here

ext/GeometryOpsLibGEOSExt/GeometryOpsLibGEOSExt.jl
```diff
@@ -3,7 +3,7 @@ module GeometryOpsLibGEOSExt
 import GeometryOps as GO, LibGEOS as LG
 import GeoInterface as GI
 
-import GeometryOps: GEOS, enforce
+import GeometryOps: GEOS, enforce, _True, _False, _booltype
```
rafaqz, line 6: Feels like we should remove the underscores if we are importing these?

src/GeometryOps.jl
```diff
@@ -2,16 +2,16 @@
...
-    # Re-export all exported symbols
-    if Base.isexported(GeometryOpsCore, name)
-        @eval export $name
-    end
-end
+import GeometryOpsCore
+import GeometryOpsCore: 
+                TraitTarget,
+                Manifold, Planar, Spherical, Geodesic,
+                BoolsAsTypes, _True, _False, _booltype,
```
rafaqz, line 9: Again the underscores on imported objects...

src/transformations/segmentize.jl
```diff
@@ -156,51 +156,64 @@ end
...
+
+# allow three-arg method as well, just in case
+segmentize(geom, max_distance::Real; threaded = _False()) = segmentize(Planar(), geom, max_distance; threaded)
+segmentize(method::Manifold, geom, max_distance::Real; threaded = _False()) = segmentize(Planar(), geom; max_distance, threaded)
+
+# generic implementation
+function segmentize(method::Manifold, geom; max_distance, threaded::Union{Bool, BoolsAsTypes} = _False())
+    @assert max_distance > 0 "`max_distance` should be positive and nonzero!  Found $(method.max_distance)."
+    _segmentize_function(geom) = _segmentize(method, geom, GI.trait(geom); max_distance)
+    return apply(_segmentize_function, TraitTarget(GI.LinearRingTrait(), GI.LineStringTrait()), geom; threaded)
```
rafaqz, line 183: Seems like code duplication with the function below, are both needed?

Are these asserts not checked twice?

(Also throwing an error is better than asserts as it's guaranteed to actually run)

src/transformations/segmentize.jl
```diff
@@ -156,51 +156,64 @@ end
...
+end
+
 function segmentize(method::SegmentizeMethod, geom; threaded::Union{Bool, BoolsAsTypes} = _False())
+    @warn "`segmentize(method::$(typeof(method)), geom) is deprecated; use `segmentize($(method isa LinearSegments ? "Planar()" : "Geodesic()"), geom; max_distance, threaded) instead!"  maxlog=3
     @assert method.max_distance > 0 "`max_distance` should be positive and nonzero!  Found $(method.max_distance)."
```
rafaqz, line 188: This assert is called again above

ext/GeometryOpsLibGEOSExt/GeometryOpsLibGEOSExt.jl
asinghvi17: Good point.  We can probably make these uppercase, but only exported from Core (not GeometryOps proper)?

rafaqz: Yep perfect

rafaqz: Just realised we can use all of these in Rasters.jl too as its all the same there already. So no underscores is good.

Really keen to unify GeometryOps/Rasters around this core for all geometry related things now.

GeometryOpsCore/src/types.jl
asinghvi17: Done

src/transformations/segmentize.jl
asinghvi17: Yes but it's a different method - you can theoretically pass an older algorithm through here, and this checks that.

PR #228, opened 2024-10-25
Fix flexi joins

asinghvi17:
I genuinely have no idea how this went undiagnosed for so long, but it turns out we weren't unwrapping the LHS in the flexi join...


rafaqz:
Prob needs another test then ;)

asinghvi17:
Will add a test in a bit but let's get this out at least :D

PR #233, opened 2024-11-15
do not export contains

alex-s-gardner:
do not export to avoid conflict with Base.contains

### DIFF
--- a/src/methods/geom_relations/contains.jl
+++ b/src/methods/geom_relations/contains.jl
@@ -1,6 +1,6 @@
 # # Contains
 
-export contains
+# do not export to avoid conflict with Base.contains
 
 #=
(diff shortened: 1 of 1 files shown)


asinghvi17 (approved):
At some point, we should look into an `@enable_base_predicates` macro, to hook geometry types into GeometryOps predicates.  But this is probably not the time to look into that.

alex-s-gardner:
> At some point, we should look into an `@enable_base_predicates` macro, to hook geometry types into GeometryOps predicates. But this is probably not the time to look into that.

Agreed. I'm not sure what @rafaqz thoughts are on that but one of the things I love about multiple dispatch is not having to do "Package.function" when the operation is intuitive as to it's purpose.

PR #238, opened 2024-11-25
Fix `overlaps` on polygons that only interact at vertices

asinghvi17:
Overlaps failed for a polygon like this (see the changes to `test` for the polygons), where the polygons overlap but only intersect at vertices (so every line-line intersection is a hinge intersection)
display
this PR attempts to move the implementation to the geom_geom_processor interface.  Still fails for lines where one line covers another.  Maybe I need some adjustments to the allows/requires?

cc @skygering


src/methods/geom_relations/overlaps.jl
```diff
@@ -76,6 +84,14 @@ overlaps(geom1, geom2)::Bool = overlaps(
     geom2,
 )
 
+
+# # Convert features to geometries
+overlaps(::GI.FeatureTrait, g1, ::Any, g2) = overlaps(GI.geometry(g1), g2)
+overlaps(::Any, g1, t2::GI.FeatureTrait, g2) = overlaps(g1, GI.geometry(g2))
+overlaps(::FeatureTrait, g1, ::FeatureTrait, g2) = overlaps(GI.geometry(g1), GI.geometry(g2))
```
asinghvi17, line 91: This is the only thing that CI doesn't cover.  We could remove it - but TBH it should be fine I think.

Issue #183, opened 2024-07-21
`rebuild` should use child geometries as the source of dimensional truth


asinghvi17:
`rebuild` takes its cue on geometry dimensions from the original geometry, not the child geometries - this means that a 2d to 3d transformation will return e.g. a 2d linestring with internal 3d points.

https://github.com/JuliaGeo/GeometryOps.jl/blob/2da44be4f3083559f06ef9a6aa7494e6c5916d3c/src/primitives.jl#L525-L534

asinghvi17:
Released in 0.1.15

PR #239, opened 2024-11-25
Check the dimensions of child geoms when rebuilding

asinghvi17:
This checks that all the child geometries have the same dimension, before rebuilding a geometry.  This enables functions like `forcexy` etc, as well as 2D to 3D transformations (or vice versa).

There is room for efficiency since we currently check all child geoms - we _could_ theoretically only check the first.  Is that something we want to do?

Fixes https://github.com/JuliaGeo/GeometryOps.jl/issues/183


rafaqz:
Looks good. Only question is the performance cost

rafaqz (approved):

asinghvi17:
I just checked this, it doubles the runtime of `GO.transform`.  If I only check the first geometry it's still 20% slower.  Maybe that's acceptable though?

src/transformations/forcedims.jl
```diff
@@ -0,0 +1,36 @@
...
+
+Force the geometry to be 3D.  Works on any geometry, vector of geometries, feature collection, or table!
+
+The `z` parameter is the default z value - if a point has no z value, it will be set to this value.  
+If it does, then the z value will be kept.
+"""
+function forcexyz(geom, z = 0)
+    return apply(GI.PointTrait(), geom) do point
+        x, y = GI.x(point), GI.y(point)
+        z = GI.is3d(geom) ? GI.z(point) : z
```
rafaqz, line 33:
```suggestion
        z = GI.is3d(geom) ? GI.z(point) : convert(typeof(x), z)
```

PR #240, opened 2024-11-25
Add a utility function to GeometryOpsCore, to get geometries from any data structure

asinghvi17:
a generic utility inspired by similar functions in Rasters.jl and elsewhere - returns a vector of the geometries in a given object, be it a feature collection, table, iterable of geometries, or whatever.

Potential next steps:
- `get_geometries(x) do geom` that rebuilds the given structure as far as possible - this is the same as `apply(::AbstractGeometryTrait)` though....
- treatment of dataframes with multiple columns

Also: do we want this here?  in GeoInterface?

### DIFF
--- a/GeometryOpsCore/src/geometry_utils.jl
+++ b/GeometryOpsCore/src/geometry_utils.jl
@@ -1,3 +1,35 @@
 
 _linearring(geom::GI.LineString) = GI.LinearRing(parent(geom); extent=geom.extent, crs=geom.crs)
 _linearring(geom::GI.LinearRing) = geom
+
+
+function get_geometries(x)
+    if GI.isgeometry(x)
+        return x
+    # elseif GI.isgeometrycollection(x)
+    #     return GI.getgeom(x)
+    elseif GI.isfeature(x)
+        return GI.geometry(x)
+    elseif GI.isfeaturecollection(x)
+        return [GI.geometry(f) for f in GI.getfeature(x)]
+    elseif Tables.istable(x) && Tables.hascolumn(x, first(GI.geometrycolumns(x)))
+        return Tables.getcolumn(x, first(GI.geometrycolumns(x)))
+    else
+        c = collect(x)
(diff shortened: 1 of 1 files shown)


GeometryOpsCore/src/geometry_utils.jl
```diff
@@ -1,3 +1,35 @@
...
+    # elseif GI.isgeometrycollection(x)
+    #     return GI.getgeom(x)
+    elseif GI.isfeature(x)
+        return GI.geometry(x)
+    elseif GI.isfeaturecollection(x)
+        return [GI.geometry(f) for f in GI.getfeature(x)]
+    elseif Tables.istable(x) && Tables.hascolumn(x, first(GI.geometrycolumns(x)))
+        return Tables.getcolumn(x, first(GI.geometrycolumns(x)))
+    else
+        c = collect(x)
```
rafaqz, line 18: Probably want to skip this if x is already an AbstractArray

rafaqz (approved):
Nice. We can delete the function from Rasters then and use this

GeometryOpsCore/src/geometry_utils.jl
```diff
@@ -1,3 +1,35 @@
 
 _linearring(geom::GI.LineString) = GI.LinearRing(parent(geom); extent=geom.extent, crs=geom.crs)
 _linearring(geom::GI.LinearRing) = geom
+
+
+function get_geometries(x)
+    if GI.isgeometry(x)
+        return x
+    # elseif GI.isgeometrycollection(x)
+    #     return GI.getgeom(x)
```
rafaqz, line 10: This would need `collect`, and needs to be checked first

asinghvi17: Checked as in?

rafaqz:
Is there a reason we didn't merge this?

asinghvi17:
Don't think so...just waiting for me to make that one change you mentioned. 

I was thinking of checking `Base.has_offset_axes` and if so calling `vec` on the array...this way we ensure that the array is one indexed.

Issue #249, opened 2025-01-11
polygonize example in docs need fix


gaelforget:
This example : https://juliageo.org/GeometryOps.jl/stable/source/methods/polygonize#Polygonizing-raster-data

This line : 

```
f, a, p = poly(polygons; label = "Polygonized polygons", axis = (; aspect = DataAspect()))
```

Gives this error : 

```
No recipe for poly with args: Tuple{GeoInterface.Wrappers.MultiPolygon{false, false, Vector{GeoInterface.Wrappers.Polygon{false, false, Vector{GeoInterface.Wrappers.LinearRing{false, false, Vector{Tuple{Float64, Float64}}, Extents.Extent{(:X, :Y), Tuple{Tuple{Float64, Float64}, Tuple{Float64, Float64}}}, Nothing}}, Extents.Extent{(:X, :Y), Tuple{Tuple{Float64, Float64}, Tuple{Float64, Float64}}}, Nothing}}, Extents.Extent{(:X, :Y), Tuple{Tuple{Float64, Float64}, Tuple{Float64, Float64}}}, Nothing}}
```

and

```
Stack trace
Here is what happened, the most recent locations are first:
error(s::String)
from 
[julia → error.jl:35](https://github.com/JuliaLang/julia/tree/8f5b7ca12ad48c6d740e058312fc8cf2bbe67848/base/error.jl#L35)
 
plot!
from 
interfaces.jl:363
connect_plot!(parent::Makie.Scene, plot::MakieCore.Poly{…}) [...show types...](http://localhost:1234/edit?id=878c0440-d004-11ef-23ce-a910af27d550#)
from 
Makie → interfaces.jl:395
 
plot!
from 
interfaces.jl:412
plot!(ax::Makie.Axis, plot::MakieCore.Poly{…}) [...show types...](http://localhost:1234/edit?id=878c0440-d004-11ef-23ce-a910af27d550#)
from 
Makie → figureplotting.jl:412
 
plot!(fa::Makie.FigureAxis, plot::MakieCore.Poly{…}) [...show types...](http://localhost:1234/edit?id=878c0440-d004-11ef-23ce-a910af27d550#)
from 
Makie → figureplotting.jl:407
 
_create_plot(F::Function, attributes::Dict{…}, args::GeoInterface.Wrappers.MultiPolygon{…}) [...show types...](http://localhost:1234/edit?id=878c0440-d004-11ef-23ce-a910af27d550#)
from 
[...]

PR #250, opened 2025-01-11
fix docs example

gaelforget:
fixes https://github.com/JuliaGeo/GeometryOps.jl/issues/249

### DIFF
--- a/src/methods/polygonize.jl
+++ b/src/methods/polygonize.jl
@@ -19,7 +19,7 @@ polygonize
 
 Here's a basic example, using the `Makie.peaks()` function.  First, let's investigate the nature of the function:
 ```@example polygonize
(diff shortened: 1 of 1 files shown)


src/methods/polygonize.jl
```diff
@@ -19,7 +19,7 @@ polygonize
 
 Here's a basic example, using the `Makie.peaks()` function.  First, let's investigate the nature of the function:
 ```@example polygonize
-using Makie, GeometryOps
+using Makie, GeometryOps, GeoInterfaceMakie, CairoMakie
```
asinghvi17, line 22:
```suggestion
using Makie, CairoMakie, GeoInterfaceMakie
using GeometryOps

PR #253, opened 2025-01-25
Allow n-dimensional arrays into polygonize in DD

asinghvi17:
This allows polygonize over an n-dimensional datacube to generate an n-2 dimensional datacube of geometries / featurecollections...kind of like R style vector data cubes.

Maybe we should allow the values to be unpacked into an n-1th dimension so we can unpack the polygons too.

### DIFF
--- a/ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
+++ b/ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
@@ -4,8 +4,10 @@ import DimensionalData as DD
 import GeometryOps as GO
 import GeoInterface as GI
 
-function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs(A), kw...)
+function GO.polygonize(A::DD.AbstractDimArray{T, N}; dims=(DD.XDim, DD.YDim), crs=GI.crs(A), kw...) where {T, N}
+    # Extract the lookups of the dimensions we care about
     lookups = DD.lookup(A, dims)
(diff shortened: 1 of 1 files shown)


ext/GeometryOpsDimensionalDataExt/GeometryOpsDimensionalDataExt.jl
```diff
@@ -14,7 +16,24 @@ function GO.polygonize(A::DD.AbstractDimArray; dims=(DD.X(), DD.Y()), crs=GI.crs
...
+    if N < 2
+        error("Array had $N dimensions, but it must have at least two!")
+    elseif N == 2
+        # If ndims == 2, then we polygonize directly, and return the output as requested.
+        Ap = PermutedDimsArray(A, dims)
+        GO.polygonize(bounds_vecs..., Ap; crs, kw...)
+    else
+        map(eachslice(A; dims = DD.otherdims(A, dims))) do a
+            ap = PermutedDimsArray(a, dims)
+            GO.polygonize(bounds_vecs..., a; crs, kw...)
```
rafaqz, line 33: Can we just pass through the dims kw instead of permuting?

asinghvi17: The x and y dims need to be in the same order as the lookups though, so I'm not sure how we'd do this...

rafaqz: Doesn't it already fix that with `dims` ordering? It probably should...

Maybe it's counter intuitive but reordering dims in the inner loop is essentially free, so usually faster than array wrappers that have a runtime Indirection

PR #257, opened 2025-02-14
Use StableTasks.jl in core

asinghvi17:
This decreases allocations by a few but no effect on runtime.  Still may make sense to add to Core.


rafaqz:
Yeah, I've noticed this being a problem elsewhere too.

rafaqz:
How maintained is StableTasks.jl?

asinghvi17:
Pretty maintained I'd say, it's a JuliaFolds2 thing so they use it in all the transducers etc.

asinghvi17:
Will replace True with TrueButStable then.

PR #258, opened 2025-02-14
Implement the GeoInterface for PolyNodes

asinghvi17:
needs tests, but this should make it so that you can plot the output of e.g. trace_polynodes.


asinghvi17:
Closing in favour of the change being made in another PR, since upstream of this is pretty outdated.

PR #269, opened 2025-03-04
[Core] Add an algorithm interface + StableTasks

asinghvi17:
This was factored out of the "dev branch" #259 and contains the subset of changes that apply to GeometryOpsCore, for easier review.

Child PRs: https://github.com/JuliaGeo/GeometryOps.jl/pull/271 (TGGeometry) -> https://github.com/JuliaGeo/GeometryOps.jl/pull/275 (AdaptivePredicates) -> #273 (clipping algorithm type) -> #274 (trees)

- Use [StableTasks.jl](https://github.com/JuliaFolds2/StableTasks.jl) in apply and applyreduce - its type-stable tasks save us some allocations!
- Remove `Base.@assume_effects` on the low level functions, which caused issues on Julia v1.11 and was probably incorrect anyway
- Add an algorithm interface with an abstract supertype `Algorithm{M <: Manifold}`, as discussed in https://github.com/JuliaGeo/GeometryOps.jl/issues/247.  Also adds an abstract Operator supertype and some discussion in code comments, but no implementation or interface surface there yet.
- Split out `types.jl` into a directory `types` with a bunch of files in it, for ease of readability / docs / use.
- (out of context change): refactor CI a bit for cleanliness.

TODOs (maybe not this PR):
- [ ] Add a `format` method that takes in an incompletely specified algorithm and some geometry as input, and returns a completely specified algorithm.  What does this mean?  Imagine I call `GO.intersection(FosterHormannClipping(), geom1, geom2)`.  That `FosterHormannClipping()` should get expanded to `FosterHormannClipping(AutoAlgorithm(), AutoAccelerator())`.  Then, `format` will take `format(alg, args...)` and:
  - get the `crstrait` of the two geometries, scan for incompatibilities, assign the correct manifold to the algorithm (maybe warn or emit debug info)
  - if no geometries available, get the manifold via `best_manifold(::Algorithm)`.
  - maybe inflate the accelerator by checking `npoint` and later preparations to see what's most efficient, maybe not - depends on what we want!

PR #271, opened 2025-03-04
Add a TGGeometry extension for fast inexact geometric predicates

asinghvi17:
Another spin out from #259

Parent: #269 
Child PRs: https://github.com/JuliaGeo/GeometryOps.jl/pull/275 (AdaptivePredicates) -> #273 (clipping algorithm type) -> #274 (trees)

TGGeometry.jl is a Julia wrapper around the [tg](https://github.com/tidwall/tg) C library for planar geometric predicates.  It doesn't use exact computation, but the indexing ideas there are extremely efficient and I thought it would be nice to at least have it accessible.

It's ridiculously fast - 100 ns for point in polygon if you convert the geometry to a TGGeometry first.

Accessible by `predicate(GO.TG(), geom1, geom2)` for all GI/Simple Features planar geometry.

This also refactors GEOS to work under the algorithm interface as well as laying the foundation for a future hypothetical PROJ algorithm, that we could use for reproject or segmentize.

> [!NOTE]
> I'm deliberately not testing TG along with GO in the operations since I don't want to deal with our tests, that require exact predicates, returning the wrong values from the inexact `tg`.  However, I should and will add some tests elsewhere.  This PR will also be a good proving ground for the new extension interface should there be one.

This is IMO ready for merge since TGGeometry.jl is tested and it doesn't impact users.  I also don't want to advertise it, at least initially.


asinghvi17:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/271?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#274** <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/274?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#273** <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/273?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#275** <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/275?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#271** <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/271?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/271?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#269** <a href="https://app.graphite.dev/github/pr/JuliaGeo/GeometryOps.jl/269?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/types.jl
```diff
@@ -33,31 +96,66 @@ Dispatch is generally carried out using the names of the keyword arguments.
...
 it must convert to the LibGEOS implementation and back - so be warned!
+
+## Extended help
+
+This uses the [LibGEOS.jl](https://github.com/JuliaGeometry/LibGEOS.jl) package,
+which is a Julia wrapper around the C library GEOS (https://trac.osgeo.org/geos).
 """
-struct GEOS
+struct GEOS <: CLibraryPlanarAlgorithm # SingleManifoldAlgorithm{Planar}
+    manifold::Planar
```
rafaqz, line 108: do we need the manifold field if its fixed to `Planar`? Or does `Planar` have fields that don't influence the type?

asinghvi17: Planar doesn't but other manifolds do.  I probably need to implement an API for this rather than implicit getproperty though...

rafaqz: Yeah, I would just make these fields the return value of an API function, relying on getproperty is fragile and having fixed type fields that do nothing is a bit confusing

asinghvi17: On the other hand, it is weird for some algorithms to store manifold but not all.  And planar is a singleton anyway so it's not like it takes extra memory or allocations...

rafaqz: Your using a field for what is actually a fixed trait.

A field communicates something that can be set to a meaningful value

src/types.jl
```diff
@@ -33,31 +96,66 @@ Dispatch is generally carried out using the names of the keyword arguments.
...
     params::NamedTuple
```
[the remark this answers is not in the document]
asinghvi17: Probably not, but I'm not sure what the effect of parametrizing these will be on compile times, since every time you change:
- kwarg order
- number of kwargs
this might force a recompilation since it's not the "same type anymore"

I guess it needs to be tested...

maybe some `@nospecializeinfer` might help, but that makes user/dev level method definitions horrible.

src/types.jl
asinghvi17: Yeah I agree that we need an API function overall.  I guess it's just a bit strange to have planar special cased as a fixed return type (all other manifolds contain information and so must be stored in the algorithm, like spherical (radius) and geodesic (ellipsoid params)).

src/types.jl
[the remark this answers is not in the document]
asinghvi17: We can figure that out later I think, if this ever becomes an issue.  (I don't think it will be since everything is passed directly to C anyway)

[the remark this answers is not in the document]
asinghvi17: Or these could hold AbstractDict or namedtuple, if compile time is an issue.

src/types.jl
```diff
@@ -23,6 +30,75 @@ useful for two reasons:
...
+Then you get a nice constructor for free, as well as the 
+`get(alg, key, value)` and `get(alg, key) do ...` syntax.
+Plus the [`enforce`](@ref) method, which will check that given keyword arguments
+are present.
+"""
+abstract type CLibraryPlanarAlgorithm <: GeometryOpsCore.SingleManifoldAlgorithm{Planar} end
+
+function (::Type{T})(; params...) where {T <: CLibraryPlanarAlgorithm}
+    nt = NamedTuple(params)
+    return T(Planar(), nt)
```
rafaqz, line 57:
```suggestion
    return T(nt)
```
This is redundant

src/types.jl
```diff
@@ -23,6 +30,75 @@ useful for two reasons:
...
+end
+(T::Type{<: CLibraryPlanarAlgorithm})(::Planar, params::NamedTuple) = T(params)
```
rafaqz, line 59:
```suggestion
(T::Type{<: CLibraryPlanarAlgorithm})(params::NamedTuple) = T(params)
```

I feel like there are some tests missing, these would error with some types

src/types.jl
```diff
@@ -23,6 +30,75 @@ useful for two reasons:
...
+
+manifold(alg::CLibraryPlanarAlgorithm) = Planar()
+best_manifold(alg::CLibraryPlanarAlgorithm, input) = Planar()
+
+function rebuild(alg::T, m::Planar) where {T <: CLibraryPlanarAlgorithm}
+    return T(m, alg.params)
+end
+
+function rebuild(alg::T, m::AutoManifold) where {T <: CLibraryPlanarAlgorithm}
+    return T(Planar(), alg.params)
```
rafaqz, line 69:
```suggestion
    return T(alg.params)
```

src/types.jl
```diff
@@ -23,6 +30,75 @@ useful for two reasons:
...
+    nt = NamedTuple(params)
+    return T(Planar(), nt)
+end
+(T::Type{<: CLibraryPlanarAlgorithm})(::Planar, params::NamedTuple) = T(params)
+
+manifold(alg::CLibraryPlanarAlgorithm) = Planar()
+best_manifold(alg::CLibraryPlanarAlgorithm, input) = Planar()
+
+function rebuild(alg::T, m::Planar) where {T <: CLibraryPlanarAlgorithm}
+    return T(m, alg.params)
```
rafaqz, line 65:
```suggestion
    return T(alg.params)
```
Again this would error for GEOS and TG

rafaqz: Again this would error in some cases

src/types.jl
asinghvi17: Resolved, we added `manifold(alg)`

rafaqz (commented):
Seems to be missing some tests?

src/types.jl
asinghvi17: I don't think it would - there's a method above that that catches (::Planar, ::NamedTuple)

But best to make the change for clarity

src/types.jl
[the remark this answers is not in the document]
asinghvi17: I guess we could also do OrderedCollections.jl's LittleDict...that would be a good middle ground at the cost of an extra ~20ms of load time (on my machine)

PR #272, opened 2025-03-04
AdaptivePredicates + minor predicate fixes

asinghvi17:
Spun out from #259

- Add AdaptivePredicates as the exact backend for orient.  At some point we should switch to an ecosystem wide (DelaunayTriangulation + GeometryOps + ...) representation for predicate kernels (Exact, Adaptive, Auto (choose adaptive if available and exact if not), and Fast).
- Fix touches for multi geometries

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Anshul Singhvi <anshulsinghvi@gmail.com>", "Rafael Schouten <rafaels
 version = "0.1.15"
 
 [deps]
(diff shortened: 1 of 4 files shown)


asinghvi17:
damn it, I should make this 2 prs...let the first one be AdaptivePredicates and minor fixes and then the second one will be the actual clipping work.

asinghvi17:
Closing in favour of other PR

PR #285, opened 2025-04-03
experimental UnitSpherical module that can express geometry on unit sphere

asinghvi17:
### TL;DR

Wrote a new `UnitSpherical.jl` module (experimental) for working with points on the unit sphere and spherical caps.

### What changed?

Created a new file `UnitSpherical.jl` that implements:

- `UnitSphericalPoint` struct for representing points on the unit sphere (S²) using Cartesian coordinates in ℝ³
- `SphericalCap` struct for representing spherical caps
- Coordinate transformations between geographic (lat/long) and unit spherical coordinates
- GeoInterface compatibility for the new types
- Utility functions for spherical geometry:
  - `spherical_distance` for calculating distances between points
  - `slerp` for spherical linear interpolation
  - `circumcenter_on_unit_sphere` for finding the center of a circle passing through three points: still broken
  - `angle_between` for calculating angles between points
  - `randsphere` for generating random points on the unit sphere

### Quick start
1. Import the module and create unit spherical points:
   ```julia
   using GeometryOps.UnitSpherical
   
   # Create a point from Cartesian coordinates
   p1 = UnitSphericalPoint(1.0, 0.0, 0.0)
   
   # Create a point from geographic coordinates (lat/long)
   p2 = UnitSphericalPoint([45.0, 90.0])
   
   # Calculate spherical distance
   dist = spherical_distance(p1, p2)
   ```

2. Test spherical caps and operations:
   ```julia
   cap1 = SphericalCap(p1, 0.5)
   cap2 = SphericalCap(p2, 0.3)
   
   # Test intersection
   _intersects(cap1, cap2)
   ```

3. Generate random points on the sphere:
   ```julia
   points = rand(UnitSphericalPoint, 100)
   ```

This should provide a nice foundation to start looking at R-trees that are backed by spherical caps, not extents, in preparation for @meggart's work involving this.  With the new changes to `as/trees` SpatialTreeInterface is now representation agnostic, meaning that as long as the predicate you passed in can consume the extent like object you have chosen, any extent like thing (accessible by `SpatialTreeInterface.node_extent`) can be used by the STI algorithms.


src/utils/UnitSpherical/cap.jl
```diff
@@ -0,0 +1,66 @@
...
+# TODO: add implementations to merge two spherical caps
+# TODO: add implementations for multitraits based on this
+
+# TODO: this returns an approximately antipodal point...
+
+
+# TODO: exact-predicate intersection
+# This is all inexact and thus subject to floating point error
+function _intersects(x::SphericalCap, y::SphericalCap)
+    spherical_distance(x.point, y.point) <= max(x.radius, y.radius)
```
meggart, line 28: Maybe I am completely misunderstanding, but shouldn't this be `spherical_distance(x.point, y.point) <= x.radius + y.radius`?

src/utils/UnitSpherical/point.jl
```diff
@@ -0,0 +1,163 @@
...
+0.0
+```
+"""
+spherical_distance(x::UnitSphericalPoint, y::UnitSphericalPoint) = acos(clamp(x ⋅ y, -1.0, 1.0))
+
+# ## Random points
+Random.rand(rng::Random.AbstractRNG, ::Random.SamplerType{UnitSphericalPoint}) = rand(rng, UnitSphericalPoint{Float64})
+function Random.rand(rng::Random.AbstractRNG, ::Random.SamplerType{UnitSphericalPoint{T}}) where T <: Number
+    θ = 2π * rand(rng, T)
+    ϕ = acos(2 * rand(rng, T) - 1)
```
meggart, line 129: I was a bit confused here since I think the convention is that `phi` resembles the longitudes and `theta` latitudes, so maybe the variable names should be exchanged here for better readability.

src/utils/UnitSpherical/cap.jl
asinghvi17: Ah yes, good point - must have been thinking of something else when I wrote that 😅

asinghvi17:
For this to work well, we'll probably want to port a _lot_ of the stuff from s2, like https://github.com/google/s2geometry/blob/58de4ea1e2f8a294e0c072c602c22232fd1433ad/src/s2/s1chord_angle.h - it's Apache 2.0 licensed so we can directly port it.

They have a good framework to handle this stuff, so if we can get the basics sorted then we're mostly there I believe.  That would also be a good foundation to build an s2 indexing framework in Julia, since we would have all the basic components already.

src/utils/UnitSpherical/cap.jl
```diff
@@ -0,0 +1,136 @@
...
+"Get the circumcenter of the triangle (a, b, c) on the unit sphere.  Returns a normalized 3-vector."
+function SphericalCap(a::UnitSphericalPoint, b::UnitSphericalPoint, c::UnitSphericalPoint)
+    circumcenter = circumcenter_on_unit_sphere(a, b, c)
+    circumradius = spherical_distance(a, circumcenter)
+    return SphericalCap(circumcenter, circumradius)
+end
+
+function _is_ccw_unit_sphere(v_0::S, v_c::S, v_i::S) where S <: UnitSphericalPoint
+    # checks if the smaller interior angle for the great circles connecting u-v and v-w is CCW
+    return(LinearAlgebra.dot(LinearAlgebra.cross(v_c - v_0,v_i - v_c), v_i) < 0)
```
asinghvi17, line 123: This needs to use robust_cross_product, probably.

src/utils/UnitSpherical/cap.jl
```diff
@@ -0,0 +1,136 @@
...
+    # small circle fits in big circle
+    return dist + small.radius < big.radius 
+end
+function _contains(cap::SphericalCap, point::UnitSphericalPoint)
+    spherical_distance(cap.point, point) <= cap.radius
+end
+
+
+function circumcenter_on_unit_sphere(a::UnitSphericalPoint, b::UnitSphericalPoint, c::UnitSphericalPoint)
+    LinearAlgebra.normalize(a × b + b × c + c × a)
```
asinghvi17, line 111: So does this

asinghvi17:
The next thing on my list is an iterator over all stable (no edges near 180 degrees) triangles of a linear ring, like what s2 does.

Issue #287, opened 2025-04-09
Using `reproject` in a threaded context makes Julia crash


mathieu17g:
This issue may be linked to #88 

Using `reproject` in a threaded context makes Julia crash. 
```julia
julia> versioninfo()
Julia Version 1.11.4
Commit 8561cc3d68 (2025-03-10 11:36 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 20 × 13th Gen Intel(R) Core(TM) i7-13700H
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, goldmont)
Threads: 20 default, 0 interactive, 10 GC (on 20 virtual cores)
Environment:
  JULIA_NUM_THREADS = 20
  JULIA_PKG_USE_CLI_GIT = false
  JULIA_SSL_CA_ROOTS_PATH = 
  JULIA_EDITOR = code
```
```julia
(jl_tbk2nN) pkg> st
Status `C:\...\AppData\Local\Temp\jl_tbk2nN\Project.toml`
  [150eb455] CoordinateTransformations v0.6.4
  [e9467ef8] GLMakie v0.11.3
  [68eda718] GeoFormatTypes v0.4.4
  [cf35fbd7] GeoInterface v1.4.1
  [db073c08] GeoMakie v0.7.11
  [3251bfac] GeometryOps v0.1.16
  [c94c279d] Proj v1.8.1
```
Below is some code to reproduce the crash

1. Create a vector of MuliPolygon
```julia
import GeoInterface as GI
import CoordinateTransformations as CT
import GeometryOps as GO

# Lambert-93 (EPSG:2154) center (https://epsg.io/2154)
center_l93 = (489_353.59, 6_587_552.2)

# MultiPolygon pattern creation
exterior = GI.LinearRing(GI.Point.([(0, 0), (40_000, 0), (40_000, 40_000), (0, 40_000), (0, 0)]))
hole = GI.LinearRing(GI.Point.([(10_000, 10_000), (20_000, 10_000), (20_000, 20_000), (10_000, 20_000), (10_000, 10_000)]))
poly1 = GI.Polygon([exterior, hole])
poly2 = GO.transform(CT.Translation(50_000, 0), poly1) # Translate the polygon to the right by 50km
mpoly = GI.MultiPolygon([poly1, poly2])

[...]

PR #288, opened 2025-04-10
Thread-safe reproject with `ThreadFunctors`

rafaqz:
Havent run this yet, but should be close to working.

@asinghvi17 turns out it needs 2 layers of functors to deal with `is3d`

Discussed in #287

Changelog:
- Refactor the `apply` pipeline to get rid of closures in favour of `Applicator`s, which are able to run `apply` on arrays, geoms and featurecollections.
- Add a `TaskFunctors` struct that allows one to use individual functors per task.  This allows thread safety for reentrant C APIs.
- Refactor `reproject`, in the Proj backend, to use this functionality


ext/GeometryOpsProjExt/reproject.jl
```diff
@@ -22,29 +22,51 @@ function reproject(geom;
...
+    threaded=False(),
     always_xy=true,
     transform=nothing,
     kw...
 )
-    transform = if isnothing(transform) 
-        s = source_crs isa Proj.CRS ? source_crs : convert(String, source_crs)
-        t = target_crs isa Proj.CRS ? target_crs : convert(String, target_crs)
-        Proj.Transformation(s, t; always_xy)
+    if istrue(threaded)
```
asinghvi17, line 42: I don't remember this function :D 

```suggestion
    if booltype(threaded) isa True

**Comment by asinghvi17** [comment]
Lol - I had a very similar approach, including an equivalent of the ApplyToN wrapper, but using TaskLocalValues.jl.  But if we never reuse the value in a task then we don't need that package.  

My struct was a bit different:
```julia

struct KnownInputDimTransformation{Z, M, T}
    transf::T
end

function KnownInputDimTransformation{Z, M}(transf::T) where {Z, M, T}
    KnownInputDimTransformation{Z, M, T}(transf)
end

function (t::KnownInputDimTransformation{Z, M, T})(p) where {Z, M, T}
    if Z
        return t.transf(GI.x(p), GI.y(p), GI.z(p))
    else
        return t.transf(GI.x(p), GI.y(p))
    end
end
```

currently does not handle Z and M but could be trivially extended to handle those and add default values if we want.

ext/GeometryOpsProjExt/reproject.jl
rafaqz: Oh I have it in Rasters I think! Should add it here it's cleaner

asinghvi17: Yeah, that would be another PR though

rafaqz: It's 4 lines ;)

asinghvi17: Well then we would go through and refactor every use of `boolastype isa True` which would take a while

asinghvi17:
Boom

```julia-repl
julia> @be GO.reproject($mpolys, $trans)
Benchmark: 13 samples with 1 evaluation
 min    7.543 ms (75012 allocs: 3.300 MiB)
 median 7.617 ms (75012 allocs: 3.300 MiB)
 mean   7.975 ms (75012 allocs: 3.300 MiB, 2.72% gc time)
 max    11.873 ms (75012 allocs: 3.300 MiB, 35.34% gc time)

julia> @be GO.reproject($mpolys, $trans; threaded = true)
Benchmark: 45 samples with 1 evaluation
 min    1.934 ms (75275 allocs: 3.496 MiB)
 median 1.970 ms (75275 allocs: 3.496 MiB)
 mean   2.229 ms (75275 allocs: 3.496 MiB, 5.59% gc time)
 max    6.006 ms (75275 allocs: 3.496 MiB, 65.74% gc time)
```

ext/GeometryOpsProjExt/reproject.jl
```diff
@@ -55,6 +59,6 @@ function reproject(geom, transform::Proj.Transformation; time=Inf, threaded = Fa
         transforms = ThreadFunctors(functors, tasks_per_thread)
         return apply(transforms, GI.PointTrait(), geom; crs=target_crs, kw...)
     else # threaded isa False
-        return apply(ApplyToPoint{_is3d(geom)}(transform), GI.PointTrait(), geom; time, target_crs, kw...)
+        return apply(ApplyToPoint{_is3d(geom)}(transform), GI.PointTrait(), geom; crs = target_crs, kw...)
```
rafaqz, line 62: I think it may be better to still make an if else block for these, putting _is3d in the type directly may still mean instability even though there are only two options

asinghvi17: can do then

GeometryOpsCore/src/types/applicators.jl
```diff
@@ -0,0 +1,105 @@
...
+struct ApplyToPoint{Z,M,F} <: Applicator{F,Nothing} 
+    f::F
+end
+ApplyToPoint{Z,M}(f::F) where {Z,M,F} = ApplyToPoint{Z,M,F}(f)
+ApplyToPoint{Z}(f::F) where {Z,F} = ApplyToPoint{Z,false,F}(f)
+# Default function is just `tuple`
+(a::Type{<:ApplyToPoint})() = a(tuple)
+
+# Currently we ignore M by default
+const ToXY = ApplyToPoint{false}
```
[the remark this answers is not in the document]
rafaqz: Yeah that's better!

PR #289, opened 2025-04-10
TaskFunctors: `length(functors) == ntasks`

asinghvi17:
small refactor

@rafaqz


rafaqz:
Added directly in the other PR

PR #305, opened 2025-05-01
Implement an Applicator WithTrait

asinghvi17:
This lets the trait, known by apply anyway, pass into the target function.  It's mostly a convenience thing but can cut down on some allocs.  Centroid does get 1000 less allocs from this, but the other improvements I made to it outshine it.

Changelog:
- Introduce a new Applicator `WithTrait((trait, obj; kw...) -> ...)` (but it usually won't get any kw, I filter the standard ones out)
- Improve the performance of area and centroid

This is really useful in fix and prepare.  They are kind of the same thing but I am going to prototype the new fix pipeline in prepare.  It's really needed, ironically, for geom-geom distance queries.  Those are n^2 unless you can do some form of branch and bound on spatial trees of the geom or rings.


rafaqz:
This is great. But my main comment is that `apply(ApplyWithTrait` is a bit confusing, to me `apply(WithTrait(f)` is cleaner language.

rafaqz (approved):

PR #312, opened 2025-05-04
add a DataFrames extension for better table reconstruction

asinghvi17:
@evetion this is what I was thinking, I benchmarked it and it does make some amount of difference, especially in the amount of memory allocated.

```julia

julia> @be GO.transform($identity, $df; copycols = $true)
Benchmark: 265 samples with 1 evaluation
 min    254.334 μs (1941 allocs: 440.797 KiB)
 median 275.834 μs (1941 allocs: 440.797 KiB)
 mean   335.600 μs (1941 allocs: 440.797 KiB, 3.15% gc time)
 max    3.747 ms (1941 allocs: 440.797 KiB, 90.64% gc time)

julia> @be GO.transform($identity, $df; copycols = $false)
Benchmark: 191 samples with 1 evaluation
 min    243.000 μs (1603 allocs: 190.797 KiB)
 median 253.583 μs (1603 allocs: 190.797 KiB)
 mean   502.920 μs (1603 allocs: 190.797 KiB, 1.03% gc time)
 max    29.574 ms (1603 allocs: 190.797 KiB, 98.68% gc time)
```

If this makes sense to you we can remove the code you have in GeoDataFrames for reproject on dataframes, only defining `reproject!`.

### DIFF
--- a/GeometryOpsCore/src/apply.jl
+++ b/GeometryOpsCore/src/apply.jl
@@ -175,9 +175,11 @@ with the same schema, but with the new geometry column.
 This new table may be of the same type as the old one iff `Tables.materializer` is defined for 
 that table.  If not, then a `NamedTuple` is returned.
 =#
-function _apply_table(f::F, target, iterable::IterableType; geometrycolumn = nothing, preserve_default_metadata = false, threaded, kw...) where {F, IterableType}
-    _get_col_pair(colname) = colname => Tables.getcolumn(iterable, colname)
-    # We extract the geometry column and run `apply` on it.
+function _apply_table(f::F, target, iterable::IterableType; geometrycolumn = nothing, preserve_default_metadata = false, threaded, kw...) where {F, IterableType}    # We extract the geometry column and run `apply` on it.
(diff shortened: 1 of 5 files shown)


asinghvi17:
@rafaqz mind giving this a look before merge?  From my perspective this is ready to go and will make GDF things much nicer.

GeometryOpsCore/src/apply.jl
```diff
@@ -246,6 +248,51 @@ function _apply_table(f::F, target, iterable::IterableType; geometrycolumn = not
     return result
 end
 
+
+"""
+    used_reconstruct_table_kwargs(input)
```
rafaqz, line 253:
```suggestion
    required_reconstruct_table_kwargs(input)
```

?

rafaqz (approved):

GeometryOpsCore/src/apply.jl
asinghvi17: They aren't required though, they all have defaults

PR #319, opened 2025-05-15
Create a specific exception type for `enforce`

asinghvi17:


GeometryOpsCore/src/types/exceptions.jl
```diff
@@ -46,4 +54,60 @@ function Base.showerror(io::IO, e::WrongManifoldException{I,D,A}) where {I,D,A}
...
+
+"""
+    MissingKeywordInAlgorithmException{Alg, F} <: Exception
+
+An error type which is thrown when a keyword argument is missing from an algorithm.
+
+The `alg` argument is the algorithm struct, and the `keyword` argument is the keyword
+that was missing.
+
+This error message is used in the [`enforce`](@ref) method.
```
asinghvi17, line 68:
```suggestion
This error message is used in the [`enforce`](@ref GeometryOps.enforce) method.
```

PR #347, opened 2025-10-18
allow table rows in relations, and reorganise boilerplate

rafaqz:
This PR lets you do this with any table:
```julia
subset = filter(df) do row
    GO.intersects(row, region) # Look, no .geometry
end
```
Or, even cleaner:
```julia
subset = filter(GO.intersects(region), df)
```

On the condition it has a `:geometry` property. Maybe we can make that `GI.geometrycolumns` ? but I'm not sure what the metadata story is for table rows. 

To avoid boilerplate I put this in an eval loop for all the methods, and noticed that Extents and Features were the same boilerplate thing, so put them in the eval loop too.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,7 +6,6 @@ version = "0.1.29"
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 AdaptivePredicates = "35492f91-a3bd-45ad-95db-fcad7dcfedb7"
-ArchGDAL = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
 CoordinateTransformations = "150eb455-5306-5404-9cee-2592286d6298"
 DataAPI = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 DelaunayTriangulation = "927a84f5-c5f4-47a5-9785-b46e178433df"


src/methods/geom_relations/common.jl
```diff
@@ -0,0 +1,28 @@
+# Identical boilerplate methods for geom relations live here
+
+for f in (:coveredby, :crosses, :disjoint, :overlaps, :touches, :within)
+    _f = Symbol(:_, f)
+
+    # Features
+    @eval begin
+        $_f(::FeatureTrait, g1, ::Any, g2) = $f(geometry(g1), g2)
+        $_f(::Any, g1, ::FeatureTrait, g2) = $f(g1, geometry(g2))
+        $_f(::FeatureTrait, g1, ::FeatureTrait, g2) = $f(geometry(g1), geometry(g2))
```
asinghvi17: Let's forward args... and kwargs... here?  There is T = Float64 and all the kwargs like exact

src/methods/geom_relations/common.jl
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    # Table rows
+    @eval begin
+        $_f(::Nothing, g1, ::Any, g2) = $f(_geometry_or_error(g1), g2)
+        $_f(::Any, g1, ::Nothing, g2) = $f(g1, _geometry_or_error(g2))
+        $_f(::Nothing, g1, ::Nothing, g2) = 
+            $f(_geometry_or_error(g1), _geometry_or_error(g2))
```
asinghvi17: ~This will probably mess with extents.  So we'll have to make sure dispatch on those is more specific.~

But maybe we could have a geometrycolumn kwarg here?  That gets passed to _geometry_or_error?

For the two arg case we can tell the user to figure it out

src/methods/geom_relations/common.jl
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    # Extent forwarding
+    $_f(t1::GI.AbstractGeometryTrait, g1, ::Any, e::Extents.Extent) =
+        $_f(t1, g1, GI.PolygonTrait(), extent_to_polygon(e))
+    $_f(::Any, e1::Extents.Extent, t2::Any, g2) =
+        $_f(GI.PolygonTrait(), extent_to_polygon(e1), t2, g2)
+    $_f(::Any, e1::Extents.Extent, ::Any, e2::Extents.Extent) =
+        Extents.$f(e1, e2)
```
asinghvi17: Ah sweet, extents are already handled

rafaqz: We need Aqua.jl but I'm scared to try at this point

src/methods/geom_relations/common.jl
rafaqz: Ok, handled all the ambiguities and added the kw

src/methods/geom_relations/common.jl
asinghvi17: Ah true, maybe I got confused with the clipping which does have that

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.29"
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 AdaptivePredicates = "35492f91-a3bd-45ad-95db-fcad7dcfedb7"
+ArchGDAL = "c9ce4bd3-c3d5-55b8-8973-c0e20141b8c3"
```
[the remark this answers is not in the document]
asinghvi17: It's already in extras I think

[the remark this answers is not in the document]
asinghvi17:
```suggestion
```

Project.toml
```diff
@@ -38,6 +39,7 @@ GeometryOpsTGGeometryExt = "TGGeometry"
 [compat]
 AbstractTrees = "0.4"
 AdaptivePredicates = "1.2"
+ArchGDAL = "0.10.10"
```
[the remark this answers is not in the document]
rafaqz: Lol gemini was being weird

PR #348, opened 2025-10-18
Use ParallelTestRunner for tests

asinghvi17:
https://github.com/JuliaTesting/ParallelTestRunner.jl is the new hotness and also allows filtering on a specific file, very nice when developing.

But it needs some changes to how we write our tests, and TestHelpers probably needs to be a separate package in the structure.


asinghvi17:
Closing since it increases test runtimes, which is the opposite of what we want

PR #352, opened 2025-10-20
add tests for simplifying small rings

rafaqz:
accidentally pushed the actual code, but here are the tests at least...


test/transformations/simplify.jl
```diff
@@ -41,3 +41,9 @@ end
         @test all(GI.coordinates(GO.simplify($linestring; tol = 100.0)) .== lg_vals)
     end
 end
+
+@testset "Simplify small LinearRing" begin
+    lr = GI.LinearRing([(1, 1), (2, 2), (3, 3), (4, 4), (1, 1)])
+    @test GI.npoint(GO.simplify(lr, tol=10)) == 3
```
asinghvi17: This seems to give 5 points...maybe you want radial instead?

PR #355, opened 2025-10-27
Sg/voronoi

skygering:
Cleaning up and adding in the wrapper of` DelaunyTriangulation`'s `voronoi` function such that it returns `GeoInterface` polygons.

### DIFF
--- a/src/methods/voronoi.jl
+++ b/src/methods/voronoi.jl
@@ -63,7 +63,7 @@ in the same order as the input points.
 !!! warning
     This interface only computes the 2-dimensional Voronoi tessellation!
     Only clipped voronoi tesselations can be created!
-    Only `T = Float64` is guaranteed good results by underlying package DelaunayTriangulation.
+    Only `T = Float64` or `Float32` are guaranteed good results by the underlying package DelaunayTriangulation.
     
 !!! note
     The polygons are returned in the same order as the input points after flattening.


skygering:
Okay I had to add some more advanced clean up for the `clip_polygon` inputs due to this bug/requirement: https://github.com/JuliaGeometry/DelaunayTriangulation.jl/issues/229

Thoughts though? Look good to go @asinghvi17?

src/methods/voronoi.jl
```diff
@@ -0,0 +1,210 @@
...
+    clip_polygon can of several types: (1) a GeoInterface polygon, (2) a two-element tuple where the first element is a list of tuple points
+    and the second element is a list of integer indices to indicate the order of the provided points, or (3) a a two-element tuple where the
+    first element is a tuple of tuple points and the second element is a tuple of integer indices to indicate the order of the provided points
+- $CRS_KEYWORD
+- `rng`: random number generator to generating the voronoi tesselation
+
+!!! warning
+    This interface only computes the 2-dimensional Voronoi tessellation!
+    Only clipped voronoi tesselations can be created!
+    Only `T = Float64` is guaranteed good results by underlying package DelaunayTriangulation.
```
DanielVandH: > Only `T = Float64` is guaranteed good results

I think since DelaunayTriangulation v1.1 this is no longer true FYI, `Float32` is also supported as long as the `AdaptiveKernel()` is used (which is the default). Maybe I've forgotten to loosen warnings about `Float32` in the documentation or have you run into some counter-examples?

skygering: I saw that in the documentation and thus included it here. That is good to know though! I will remove this warning!

asinghvi17:
```suggestion
    Only `T = Float64` or `Float32` are guaranteed good results by the underlying package DelaunayTriangulation.
```

asinghvi17 (approved):

PR #357, opened 2025-10-30
import Wrappers names and test

rafaqz:
Just to complete the namespace with e.g. `GeometryOps.Polygon` and test these and the rest we already had from GeoInterface

### DIFF
--- a/test/primitives.jl
+++ b/test/primitives.jl
@@ -14,18 +14,18 @@ using ..TestHelpers
 
 pv1 = [(1, 2), (3, 4), (5, 6), (1, 2)]
 pv2 = [(3, 4), (5, 6), (6, 7), (3, 4)]
-lr1 = GI.LinearRing(pv1)
-lr2 =  GI.LinearRing(pv2)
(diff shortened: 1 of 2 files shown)


test/primitives.jl
```diff
@@ -14,18 +14,18 @@ using ..TestHelpers
...
+lr2 =  GO.LinearRing(pv2)
+poly = GO.Polygon([lr1, lr2])
 
 @testset "apply" begin
 
     @testset_implementations "apply flip" begin
-        flipped_poly = GO.apply(GI.PointTrait, $poly) do p
-            (GI.y(p), GI.x(p))
+        flipped_poly = GO.apply(GO.PointTrait, $poly) do p
+            (GO.y(p), GO.x(p))
```
asinghvi17: Not sure this should be GO and not GI...I can see the argument for consistency though

asinghvi17: that should be what Geo is for though.  I don't want to advertise `GO.x` because then people will try to override it.

PR #359, opened 2025-11-06
Add badge for blazingly fast performance

asinghvi17:


rafaqz:
Haha Rasters needs that too

PR #366, opened 2025-12-13
Spherical area using Girard's method

asinghvi17:
This is a bit bad for basically any edge case, but for a reasonably sized quadrilateral, which is what I really care about right now, this is fine.  The next PR I'll have Claude port out s2's algorithm to split the polygon into triangles that are numerically stable.


asinghvi17:
There's some kind of bug in this - it gets the area of Switzerland wrong

PR #383, opened 2026-01-19
Add rectangletrait dispatch for Extent.

evetion:
To handle https://github.com/JuliaGeo/GeoInterface.jl/pull/211 in a backwards compatible manner.


evetion:
I recently hit this again, can we merge it?

PR #395, opened 2026-03-31
Pass `exact` through point intersections

Maximus-08:
## Summary

- expose the `exact` keyword on public `intersection_points` entrypoints
- pass `exact` through `intersection(...; target = GI.PointTrait())` to `_intersection_points`
- add regression tests for both public point-intersection paths with `exact = false`

Before this change:
- `_intersection_points(...; exact=...)` already supported the keyword
- `intersection_points(...)` dropped it
- `intersection(...; target = GI.PointTrait())` also dropped it

That meant users could not control the exact/inexact predicate path from the supported public APIs.

## Testing
- full package test suite passed

Documenter errors like CurrentModule = GeometryOpsCore failing because GeometryOpsCore was not defined in Main during doctest evaluation. This error is also seen in the CI doctest

### DIFF
--- a/src/methods/clipping/intersection.jl
+++ b/src/methods/clipping/intersection.jl
@@ -40,11 +40,11 @@ GI.coordinates.(inter_points)
 ```
 """
 function intersection(
-    alg::FosterHormannClipping, geom_a, geom_b, ::Type{T}=Float64; target=nothing, kwargs...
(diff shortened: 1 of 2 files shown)


Maximus-08:
Hey @asinghvi17 can you please take a look when you have time, all the tests pass, I added some tests to increase the code coverage

PR #406, opened 2026-05-29
Implements min_number for segmentation

### LINKED ISSUE #405: Add min_number parameter to segmentize
I am working on a neural network which operates on polylines with variable number of vertices. To fit each polyline into the ANN the simplify method works well to reduce the number of vertices of long polylines. However to *increase* the number of vertices for short polylines, I am out of luck (currently I just fill the matrix with zeros but I suspect that may limiting the sensitivity of the model).

I am thinking that `segmentize` could work similar to `simplify` where you either pass in `max_distance` or `min_number` (but not both), and if you pass in `min_number` a simple algorithm of inserting a redundant vertex in the center of the longest line iteratively until the `min_number` is reached.

runarberg:
This is implemented with a simple algorithm:

- Compute all distances between subsequent vertices (segments) in the geometry.
- While the number of vertices is less then the desired number:
  - Find the longest (max₁) and the second longest (max₂) segments.
  - Find lowest number (n > 1) such that max₁ can be split into evenly into n segments which are shorter then max₂.
  - Split max₁ into n equal segments.
  - Add the new vertices to the geometry.
  - Update the distance array with the new distances.

resolves #405

### DIFF
--- a/src/transformations/segmentize.jl
+++ b/src/transformations/segmentize.jl
@@ -4,11 +4,12 @@ export segmentize
 export LinearSegments, GeodesicSegments
 
 #=
-This function "segmentizes" or "densifies" a geometry by adding 
-extra vertices to the geometry so that no segment is longer than 
-a given distance.  This is useful for plotting geometries with a 
-limited number of vertices, or for ensuring that a geometry is not 
-too "coarse" for a given application.
+This function "segmentizes" or "densifies" a geometry by adding extra
+vertices to the geometry so that either a) no segment is longer than a
+given distance, or b) total number of vertices is no less then a given
+number. This is useful for plotting geometries with a limited number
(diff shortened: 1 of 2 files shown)


runarberg:
I change the algorithm a little from what I originally proposed in #405. The benefits of this one is it is a little more optimized (will require fewer iterations) and it will add vertices in very similar places as `max_distance` (I think they start to differ when `min_number` revisits a segment it has split before).

I also tried to make sure to retain the inner structure of the geometry (e.g. if using Float32). Other methods don’t do that, but I felt it was worth it to try it. I am using Float32 in my model but `simplify` will change my vertices to Float64. If you are fine doing it this way I may create a separate issue to this in more places, including `simplify`.

PR #411, opened 2026-06-09
Add separating-axis early-out to ConvexConvexSutherlandHodgman for disjoint inputs

### LINKED ISSUE #408: ConvexConvexSutherlandHodgman: separating-great-circle early-out for disjoint inputs
`intersection(ConvexConvexSutherlandHodgman, a, b; target=PolygonTrait())` runs the full clip and allocates intermediate point buffers even when `a` and `b` are disjoint, only to return an empty result.

This is hot downstream: in ConservativeRegridding.jl the spatial-tree candidate pairs are filtered on loose `SphericalCap` overlap, so ~80% of clipped pairs turn out disjoint/edge-touching (exactly zero area) — each paying a full clip (~29 allocations) for nothing.

**Recommendation: a separating-great-circle early-out.** For each edge of each convex polygon, the plane through the origin spanned by that edge's two endpoints is a candidate separating plane; if every vertex of the other polygon lies strictly on the far side, the polygons are disjoint → return empty with no allocation. O(n+m), allocation-free.

(GeometryOps v0.1.40.)

asinghvi17:
Fixes #408.

`intersection(ConvexConvexSutherlandHodgman(...), a, b)` previously ran the full Sutherland-Hodgman clip — allocating intermediate point buffers — even when the two convex polygons are disjoint, which is the dominant case in workloads like ConservativeRegridding.jl (~80% of spatial-tree candidate pairs).

## Change

Adds a separating-axis early-out before any buffers are built, for both manifolds:

- **Spherical**: for each edge of either polygon, the edge's great-circle normal (`robust_cross_product`, hoisted out of the vertex loop) defines a candidate separating plane. If every vertex of the other polygon lies strictly on the negative side (semantics exactly matching `spherical_orient < 0`, same tolerance), the polygons are disjoint and the empty degenerate polygon is returned immediately.
- **Planar**: the classic SAT equivalent using `Predicates.orient`.

The check is only a *sufficient* condition: touching, shared-edge, and exactly-antipodal configurations (where vertices lie on the candidate great circles) fall through to the full clip, so results never change — only speed. Degenerate edges (e.g. the ring's closing point) are skipped, since `robust_cross_product(a, a)` returns an arbitrary orthogonal vector that must not be used as a candidate separating plane.

## Type-level toggle

The early-out is selected at the type level via a new `DisjointCheck` enum carried as a second type parameter, defaulting to on:

```julia
ConvexConvexSutherlandHodgman()                                # ::ConvexConvexSutherlandHodgman{Planar, CheckDisjoint}
ConvexConvexSutherlandHodgman(Spherical(), SkipDisjointCheck)  # early-out disabled
```

Since the enum value is a type parameter, the branch is resolved at compile time. Existing dispatch on `ConvexConvexSutherlandHodgman{Planar}` etc. still works via partial instantiation, and the constructor validates the parameter is a `DisjointCheck` value.

## Benchmarks (Chairmarks, min time, 1°×1° cells)

| Case | SkipDisjointCheck (= before) | CheckDisjoint (default) |
|---|---|---|
| spherical disjoint | 367 ns, 24 allocs | **40 ns, 6 allocs (~10×)** |
| planar disjoint | 60 ns, 10 allocs | **25 ns, 6 allocs (~2.4×)** |
| spherical adjacent (shared edge) | 383 ns | 398 ns (falls through) |
| spherical overlapping | 431 ns | 495 ns (cost of the failed sweep) |
| planar overlapping | 170 ns | 166 ns (unchanged) |

The remaining 6 allocations on the disjoint path are the returned degenerate polygon itself. For the ~80%-disjoint regridding mix described in the issue this is roughly a 3× overall speedup of the clipping step; workloads where pairs almost always overlap can opt out with `SkipDisjointCheck`.

## Tests

- Direct predicate tests for both helpers: disjoint far/near/diagonal-gap pairs trigger in both argument orders; shared-edge, corner-touching, overlapping, and identical inputs do not (proving the conservative fall-through).
- End-to-end zero-area results for disjoint inputs in both argument orders, plus an exactly-antipodal spherical pair documenting that the early-out stays conservative while the full clip still returns empty.
- The `DisjointCheck` toggle: default parameter value, invalid-parameter `ArgumentError`, and identical results with the check disabled on both manifolds.

### DIFF
--- a/src/methods/clipping/sutherland_hodgman.jl
+++ b/src/methods/clipping/sutherland_hodgman.jl
@@ -1,25 +1,25 @@
 # # Sutherland-Hodgman Convex-Convex Clipping
 export ConvexConvexSutherlandHodgman
-export SeparatingAxisCheck, CheckSeparatingAxis, SkipSeparatingAxisCheck
+export DisjointCheck, CheckDisjoint, SkipDisjointCheck
 
 """
-    SeparatingAxisCheck
+    DisjointCheck
 
 Enum controlling whether [`ConvexConvexSutherlandHodgman`](@ref) runs a
 separating-axis early-out for disjoint inputs before clipping. Values:
 
-- `CheckSeparatingAxis` (default): test each edge line (planar) or edge great
+- `CheckDisjoint` (default): test each edge line (planar) or edge great
   circle (spherical) as a candidate separating axis, and return the empty
   result without allocating if one is found. This makes disjoint pairs
   allocation-free at the cost of one extra O(n*m) predicate sweep on
   overlapping pairs.
-- `SkipSeparatingAxisCheck`: always run the full clip.
+- `SkipDisjointCheck`: always run the full clip.
 """
-@enum SeparatingAxisCheck CheckSeparatingAxis SkipSeparatingAxisCheck
+@enum DisjointCheck CheckDisjoint SkipDisjointCheck
 
 """
     ConvexConvexSutherlandHodgman{M <: Manifold, S} <: GeometryOpsCore.Algorithm{M}
-    ConvexConvexSutherlandHodgman(manifold = Planar(), check = CheckSeparatingAxis)
+    ConvexConvexSutherlandHodgman(manifold = Planar(), check = CheckDisjoint)
 
 Sutherland-Hodgman polygon clipping algorithm optimized for convex-convex intersection.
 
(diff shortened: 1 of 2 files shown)


asinghvi17:
This may not actually help in the use case I care about, so I'm going to leave it as a draft for now until I can establish whether it helps.

PR #428, opened 2026-07-03
[WIP] Prepared geometry, for real this time

asinghvi17:
Prepared geometry wrappers, recursive, with type based keys to extract preparations.  Only planar for now but will also try to wire spherical in.  

The main preparation right now is an edge list spatial tree, can be either a natural tree or some form of R-tree, which we have an internal, flattened implementation for.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #431, opened 2026-07-06
Add `FlexibleRTrees` and speed up `NaturalIndex` traversal

asinghvi17:
Split out of #428 so the spatial-tree improvements can land independently of prepared geometry.

## `FlexibleRTrees` (new)

A packed , static R-tree over `Extent`s of any dimensionality, vendored and restructured from SortTileRecursiveTree.jl. Storage is a flat vector of per-level extent vectors plus a leaf permutation (like `NaturalIndex` and JTS `HPRtree`), so `RTree{A, E}` is concrete regardless of depth and fully inferrable.

Benchmarks: on spatially random collections, HPR beats STR by ~20% and unsorted packing by ~1000×; on ring edges natural order remains fastest, confirming `NaturalIndex` as the default there.  The other advantage here is type stability, so you can do things like nearest neighbour queries and distance queries on it.

## `NaturalIndex` traversal speed-up

Each per-child operation (`nchild`, `getchild`, `child_indices_extents`) re-chased `parent_index → levels → level → extents` for every child, and profiles showed that pointer chase dominating tree-query time. Hoisting the child-extent vector once per node makes generic SpatialTreeInterface queries ~1.7× faster (point-in-polygon over an indexed ring: 300 ns → 178 ns per query on a 65k-vertex polygon) with no interface or layout change.

## Relation to #428

The prepared-geometry integration stays there: the `build_edge_tree` hook that lets a bulk-load algorithm select `RTree` as an edge-tree backend, and the corresponding integration testset.


src/utils/FlexibleRTrees/FlexibleRTrees.jl
```diff
@@ -0,0 +1,278 @@
...
+    end
+    return reverse!(levels)
+end
+
+# ## Extent helpers
+
+_ndims(::Type{Extents.Extent{K, V}}) where {K, V} = length(K)
+_center(ext::Extents.Extent) = map(b -> (b[1] + b[2]) / 2, values(ext))
+
+include("hilbert.jl")
```
asinghvi17: All includes need to be reasonable. Maybe "flexible arteries" just becomes a purely included and using file, and then we move the generic stuff to a `utils.jl`, or we also have a `types.jl` that would work.

PR #434, opened 2026-07-06
Add spherical arc extents and manifold-aware `Extents.extent`

asinghvi17:
Makes spherical extents first-class, in three layers. Follow-up to #432; #436 (stacked on this PR) builds `RTree`/`NaturalIndex` construction on top of it.

**`spherical_arc_extent(a, b)`** — new public API in `UnitSpherical`: the 3D Cartesian `Extent{(:X, :Y, :Z)}` of the shorter great-circle arc between two points on the unit sphere. An arc bulges out of its endpoints' bounding box (two points at `z = 0.9` joined over the pole reach `z = 1`), so unioning vertex extents is not a valid bound. Trig-free: each coordinate along the arc is a sinusoid of amplitude `hypot(aᵢ, t̂ₐᵢ)`, with an interior extremum iff the coordinate rises at `a` and falls at `b`; tangents come from `robust_cross_product` (nearly-degenerate and nearly-antipodal arcs stay stable), and a 4-ulp pad guarantees the box contains the arc.

**Manifold-aware edge decomposition** — `eachedge`, `to_edgelist`, and `lazy_edgelist` gain manifold-first methods: `Spherical()` yields edges as `UnitSphericalPoint` pairs whose `GI.Line`s carry arc extents. Rings already made of `UnitSphericalPoint`s take this path with no manifold argument (previously a `MethodError` from the 2D-hardcoded `_lineedge`).

**`Extents.extent(m::Manifold, geom, [T])`** — reachable as `GO.extent` (GO does not export `extent`; the methods extend `Extents.extent`). `Planar()` delegates to `GI.extent`. On `Spherical()`, curves are unions of edge arc extents; rings and polygons are regions under S2's loop convention — CCW, interior on the left — so a CW ring encloses the complement. A region's coordinate extremum lies on its boundary or at one of the six axis points `(±1,0,0), (0,±1,0), (0,0,±1)` (every polar lonlat cell covers a pole), so each axis point gets an enclosure check by S2-style crossing parity. Degenerate anchor edges are retried; if every anchor is degenerate the axis extends to `±1` — possibly loose, never under-covering.

## Also fixed

`to_unit_spherical_points` reinterpreted the first two Cartesian coordinates of `UnitSphericalPoint` input as geographic lon/lat; it now passes such points through, as its docstring claimed.

## Tests (~1,000 new assertions)

[...]

PR #438, opened 2026-07-13
Unify the spherical relate kernel with the shared extent substrate

asinghvi17:
Stacked on #422 (base branch `relateng-spherical`), which stacks on #416. **Note for review:** the branch merges `spherical-arc-extent` (#434) to get the shared substrate, so the three-dot diff includes #434's content until it lands on `main`; the changes to review are the four commits after the merge (`84e02ce…8d5ea7e`, +250/−280 net).

## What this does

Makes the spherical relate kernel *consume* the shared spherical substrate (#434) instead of carrying a private fork of it. The kernel keeps exactly what is relate-specific — point conversion, ingest validation, exact predicates, padding — and delegates the geometry.

**Extent delegation.** `rk_interaction_bounds(::Spherical)` now composes `spherical_arc_extent` per edge and `_spherical_region_extent` for area interiors, over kernel-converted points (so boxes and ingested vertices agree bit-for-bit); the kernel's own `arc_extent`, `_point_box`, `_arcs_extent`, `_sph_bounds`, and `_widen_area_axes` are deleted. Area widening thereby upgrades from the winding-sum test (documented super-hemisphere limitation) to the shared crossing parity with anchor retry and conservative ±1 fallback. Two deliberate seams:

- Rings keep **dim-1 curve bounds** (JTS envelope semantics), not S2 region semantics — a CW hole must not become a complement region.
- The **antipodal-edge throw** moves from the extent computation to an explicit ingest hook (`_validate_relate_edges`, run once per curve at `RelateGeometry` construction), since `spherical_arc_extent` deliberately picks a stable plane instead of throwing.

**Ring containment.** The anchor-retry crossing parity from #434 is promoted to `UnitSpherical.spherical_ring_contains` (documented, exported) with its three geometric predicates injectable — `orient`, `on_arc`, `proper_crossing` — defaulting to the tolerance-based substrate predicates, so the extent path is unchanged. `rk_point_in_ring(::Spherical)` keeps its exact boundary classification and injects `rk_orient` + `_arcs_cross_properly` for an interior decision as exact as the kernel's predicates. This retires the fixed north-pole reference walk and its documented degeneracies (query at/antipodal to the pole, vertex on the query meridian); an all-anchors-degenerate configuration now throws informatively instead of answering wrong.

**Pruning.** `rk_bounds_disjoint` (= `!Extents.intersects` renamed) and `rk_bounds_covers` (no callers) are deleted; extent tests use Extents.jl vocabulary directly. Stored-extent reuse on `Spherical` now requires a 3D `(X, Y, Z)` extent, so a user's lon/lat box is never compared against unit-sphere boxes, and `_union_stored_extents` computes missing child extents in the manifold's coordinate space.

## Tests

Full suite green: JTS XML conformance 6,537; RelateNG engine 1,868; LibGEOS differential fuzz 500; spherical kernel conformance 51,614 (the arc-extent containment testset moved to `test/utils/unitspherical.jl` with the substrate); spherical end-to-end incl. the relocated antipodal guard; allocations + type stability.


asinghvi17:

PR #439, opened 2026-07-13
Prune implicit machinery flagged by the substrate audit

asinghvi17:
Stacked on #438 (base branch `relateng-shared-substrate`). The four small cleanups from the substrate audit; each is its own commit.

- **Merge the byte-identical `Planar`/`Spherical` twin methods.** `_select_edge_set_accelerator` and the `AutoAccelerator` arm of `_build_prepared_edge_index` now dispatch on `Union{Planar, Spherical}` (the manifolds with a segment-extent kernel). Also fixes the `AutoAccelerator` docstring, which still claimed non-`Planar` manifolds never take the tree path.

- **Name the relateng tree accelerator `DoubleNaturalTree`.** The tree path builds a `NaturalIndex` over each side's segment extents and traverses the two simultaneously — exactly what clipping's existing `DoubleNaturalTree` names. `AutoAccelerator` selected (and the docs advertised) `DoubleSTRtree`, which builds no STRtree here.

- **Remove the lazy point-in-area auto-index.** Unprepared `locate_on_polygonal` counted queries per polygonal element and silently switched to a lazily built `IndexedPointInAreaLocator` on the 9th — an implicit adaptive index deviating from Java, which keys indexing purely on `isPrepared`. Restored the Java rule: unprepared mode scans the rings on every query; repeated point location against one geometry is what `prepare` (the explicit escape hatch) is for. This orphans `sort_leaves = false`, so `SortedPackedIntervalRTree` always midpoint-sorts as JTS does (the measured sorted-vs-ring-order tradeoff stays recorded in its docstring), and `poly_query_count` / `_LAZY_INDEX_QUERY_THRESHOLD` are deleted.

  For the record, the removed heuristic's rationale: an unsorted index build cost ~10–13 ring scans (Natural Earth coastlines), so switching at 8 queries had worst-case regret of about one build. If unprepared multi-element relates ever show up hot, `prepare` covers it explicitly.

- **Delete the unused `is_known_entry`.** As ported it could never return `false` (matrix entries start at DIM_FALSE and only increase, so DIM_UNKNOWN never appears); a comment keeps the JTS-parity gap explained.

## Tests

Full RelateNG suite green (JTS XML conformance, engine, LibGEOS differential fuzz, spherical conformance + end-to-end, allocations + type stability). The lazy-index testset now asserts the unprepared locator never builds an index; the prepared-vs-unprepared differential is unchanged.

### DIFF
--- a/src/methods/geom_relations/relateng/edge_intersector.jl
+++ b/src/methods/geom_relations/relateng/edge_intersector.jl
@@ -180,9 +180,9 @@ recorded on `computer`.
 - `NestedLoop`: a plain double loop over
   string pairs and segment pairs, with a per-pair segment-extent
   disjointness skip (on `Planar`).
-- Any tree-backed accelerator (e.g. `DoubleSTRtree`): a spatial index
-  (`_relate_edge_index`, currently a `NaturalIndex`) is built over the
-  per-segment extents of each side and traversed with
+- Any tree-backed accelerator (canonically `DoubleNaturalTree`): a spatial
+  index (`_relate_edge_index`, currently a `NaturalIndex`) is built over
+  the per-segment extents of each side and traversed with
   `SpatialTreeInterface.dual_depth_first_search`
   under the `Extents.intersects` predicate.
 - [`AutoAccelerator`](@ref): picks `NestedLoop` below the clipping size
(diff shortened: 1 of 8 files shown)


asinghvi17:

PR #440, opened 2026-07-13
Carry payloads in `RTree` and dissolve the relate edge-index side-tables

asinghvi17:
Stacked on #439 (base branch `relateng-cleanups`). The "payload-carrying tree" follow-up from the substrate audit, in two commits.

**`RTree` carries the collection it indexes.** Queries already return indices into the input collection, but the tree threw the collection away — so every consumer had to carry a parallel payload vector alongside the tree (JTS trees carry their items; the index-returning design needs the vector to live *somewhere*). It now lives on the tree as `tree.data`. Two supporting changes:

- An `extents` kwarg supplies precomputed extents (one per element, in order) for payload elements that carry no extent of their own, or extents computed in another coordinate space — the relate kernel computes segment extents over kernel-converted points, which `GI.extent` on the payload could never reproduce.
- `Unsorted` loading returns `Base.OneTo` from `loadorder`, so the constructor skips the identity permutation and aliases the extents vector as the leaf level. Zero copies — the same build cost as `NaturalIndex`, making `RTree(Unsorted(), ...)` a true natural index. (`RTreeNode` is reparametrized on the concrete tree type since `RTree` gained type parameters.)

**The relate edge index dissolves its side-tables.** `_relate_edge_index(m, ss_list)` now returns a natural-order `RTree` whose data is each leaf's (string index, segment index) owner, built from the flattened segment extents (or `nothing` for empty lists). The three dual-traversal sites map hits through `tree.data[i]` instead of a parallel `owners` vector, and `PreparedEdgeIndex` — a struct pairing exactly that tree with exactly that vector — dissolves into the tree itself: `PreparedRelate.edge_tree` is now the `RTree` or `nothing`.

This sets up the interval-tree follow-up: `SortedPackedIntervalRTree` is a midpoint-sorted packed tree over 1-D intervals carrying segment payloads, which is `RTree(STR(), segs; extents = y_intervals)` — STR in one dimension *is* the midpoint sort.

## Tests

New `FlexibleRTrees` testset covers payload round-trips under reordering, the `Unsorted` zero-copy aliasing, collected non-vector input, and the extents/data length mismatch throw; the existing brute-force query sweeps now exercise the zero-copy `Unsorted` path. RelateNG: engine + prepared mode, edge intersector accelerator differentials, spherical end-to-end, JTS XML conformance, LibGEOS fuzz, and the allocation budget (256-vertex pairs exceed the tree threshold, so the budget covers the new build path) all green.

### DIFF
--- a/src/methods/geom_relations/relateng/edge_intersector.jl
+++ b/src/methods/geom_relations/relateng/edge_intersector.jl
@@ -181,9 +181,9 @@ recorded on `computer`.
   string pairs and segment pairs, with a per-pair segment-extent
   disjointness skip (on `Planar`).
 - Any tree-backed accelerator (canonically `DoubleNaturalTree`): a spatial
-  index (`_relate_edge_index`, currently a `NaturalIndex`) is built over
-  the per-segment extents of each side and traversed with
-  `SpatialTreeInterface.dual_depth_first_search`
+  index (`_relate_edge_index`, a natural-order `RTree` carrying segment
+  owners as data) is built over the per-segment extents of each side and
+  traversed with `SpatialTreeInterface.dual_depth_first_search`
   under the `Extents.intersects` predicate.
 - [`AutoAccelerator`](@ref): picks `NestedLoop` below the clipping size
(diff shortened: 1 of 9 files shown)


asinghvi17:

PR #441, opened 2026-07-13
Re-express the point-in-area interval index on the shared `RTree`

asinghvi17:
Stacked on #440 (base branch `relateng-payload-tree`) — the last of the three substrate-audit follow-ups.

`SortedPackedIntervalRTree` was a from-scratch packed tree that duplicated the shared bulk-loading substrate piece by piece: its midpoint sort (JTS `NodeComparator`) is exactly sort-tile-recursive in one dimension, and its flat per-level min/max arrays are the same bottom-up consecutive-run extent packing. With #440's payload-carrying `RTree`, the whole thing is expressible as a constructor call: the prepared point-in-area locator's segment index is now

```julia
RTree(STR(), segs; extents = y_intervals)   # 1-D (Y,) extents
```

queried through `SpatialTreeInterface.depth_first_search` with a closed `[y, y]` interval extent (the do-block callback form, so the RayCrossingCounter loop stays allocation-free). What's preserved and what changed:

- The **1-D stabbing contract is unchanged**: the horizontal ray must visit every segment whose y-interval contains `p.y` regardless of x, so a 2-D index could not prune more without changing the ray-counting contract; closed-interval touch semantics match (`Extents.intersects` is closed).
- The **midpoint-sorted layout is preserved by construction** — STR's first (only) dimension sort *is* the midpoint sort, keeping the measured ~3× query win over ring order for this build-once-query-forever index.
- The **tree shape changes** from JTS's binary pairwise packing to the `RTree` default 16-ary fanout; query results are extent-checked and set-identical (the test suite compares against brute force).
- `IntervalIndexedGeometry.isEmpty` becomes `index === nothing` (an empty polygonal geometry contributes no segments), replacing the separate `is_empty` flag.

Net: −104 lines (the source file shrinks by ~145), and RelateNG's last private index structure now sits on the shared substrate.

## Tests

The `SortedPackedIntervalRTree` testset is re-expressed as a 1-D stabbing testset over the new index shape (same cases: duplicates, point intervals, closed-interval touches, brute-force sweeps). Green on this branch: indexed point-in-area (incl. the 10k-vertex prepared-vs-unprepared differential), point locator, allocations + type stability, LibGEOS differential fuzz (500).

⚠️ The two long gate files (`relate_ng.jl` engine, `xml_suite.jl` JTS conformance) were repeatedly terminated externally while running in the background (the second engine attempt died ~19 min in, likely moments from passing); they passed on the base branch #440 and this branch touches only the prepared locator's index internals, which the fuzz + differential testsets exercise end-to-end. They still need one clean confirmation run before undrafting.

### DIFF
--- a/src/methods/geom_relations/relateng/indexed_point_in_area.jl
+++ b/src/methods/geom_relations/relateng/indexed_point_in_area.jl
@@ -1,140 +1,30 @@
 # # RelateNG indexed point-in-area location
 #
 # Prepared-mode point-in-area locator for RelateNG (Task 22). This file holds
-# the ports of the three JTS classes behind prepared-mode point location, in
-# this order (JTS file boundaries preserved as clearly marked sections):
+# the ports of the two JTS classes behind prepared-mode point location
+# (JTS file boundaries preserved as clearly marked sections):
 #
-# 1. `SortedPackedIntervalRTree`  (JTS index/intervalrtree/SortedPackedIntervalRTree.java)
-# 2. `RayCrossingCounter`         (JTS algorithm/RayCrossingCounter.java)
-# 3. `IndexedPointInAreaLocator`  (JTS algorithm/locate/IndexedPointInAreaLocator.java)
+# 1. `RayCrossingCounter`         (JTS algorithm/RayCrossingCounter.java)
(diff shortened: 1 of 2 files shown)


asinghvi17:

PR #442, opened 2026-07-16
Add the OverlayNG noding substrate (phase 1): exact noded arrangements with symbolic nodes

asinghvi17:
Phase 1 of the OverlayNG port: the **noding substrate** — from two geometries to an exact noded arrangement, plus coordinate emission. Stacked on `relateng` (#416); merge after it. Design doc: `docs/plans/2026-07-16-overlayng-noding-substrate.md` (first commit on this branch), which also records the survey + spike evidence base (S1–S4) behind every decision here.

## Design in one paragraph

Overlay is computed as an **exact arrangement with symbolic nodes, rounded once at emission** — the constructive extension of RelateNG's design D2. There is no snapping, no snap-rounding, no precision model, and no retry ladder, permanently: that JTS machinery patches an inexact-noding substrate this engine doesn't have. No constructed coordinate ever enters a decision; Float64 appears only inside certified filters (which escalate to the exact kernel predicates on uncertainty) and at final emission. The pipeline keeps JTS OverlayNG's shape (node → graph → label → build) so later phases stay file-by-file diffable against Java.

## What's here

- **`src/methods/clipping/overlayng/noding/`** (~600 SLOC, all internal, nothing exported):
  - `noded_arrangement.jl` — `NodedArrangement{P}` / `NodedEdge`: noded edges are `(string, segment, node_lo, node_hi)` references; a noded edge carries no geometry, so source metadata cannot desynchronize.
  - `collect.jl` — candidate enumeration via the existing payload `RTree` + `dual_depth_first_search`, classification via `rk_classify_intersection`; proper crossings become symbolic nodes on both parents, touch/collinear incidences become input-vertex nodes (asserted, not assumed). Accepts optional prebuilt trees (a prepared geometry's edge index is the same structure).
  - `node_identity.jl` — two-tier node identity: egal `NodeKey` Dict, then a float-sweep with certified radii proposing candidate coincidences confirmed by `rk_nodes_coincide`, union-find over confirmed pairs only (zero on all real data measured; exercised by degree-6 constructed cases).
  - `split.jl` — per-segment ordered/deduped interior node lists → noded edges, zero-length guarded.
  - `emit.jl` — the only lossy step, memoized: planar crossings via a **certified double-double fast path** (accepted iff provably correctly rounded, `Rational{BigInt}` fallback); spherical crossings via a conditioning-gated float direction (exact `_sph_crossing_dir` fallback); vertex nodes pass through bit-exact.
- **Kernel substrate** (in the relateng kernel files, where manifold predicates live):
  - `rk_compare_along_segment` — new `rk_` predicate ordering nodes along a segment: certified float stage, lazy exact fallback (S1 measured the float stage resolving 100% of comparisons on real data, 240–590× vs exact-always in dense regimes).
  - `_ring_material_interior_on_left` — the material-interior orientation authority (the hole-side flip exists exactly once, next to `_ring_interior_on_left`), which phase 2's `depth_delta` derivation consumes.
  - One unified exact-crossing call shape serving ordering keys, coincidence, and emission.

## Validation

- New tests: 7,490 assertions (`test/methods/clipping/overlayng/noding.jl`) + 422 (`test/methods/relateng/kernel_along_segment.jl`): the five arrangement invariants on constructed cases (degree-6 coincident node, crossing-on-third-string-vertex, vertex-identical shared borders → zero phantom crossings, a–b–a spikes), float-vs-exact-always ordering cross-check, emission certificate audit (certified result == rational result on every node, >7,000 nodes), and an env-gated NE-110m real-data subset asserting the invariants plus the rounded-arrangement audit (zero rounding-introduced crossings, both manifolds).
- Existing kernel gates re-run at their exact counts (spherical kernel conformance ~51,943; S2 crossings 4,022).
- Performance vs the spike budgets: planar emission 0.026 µs/node (budget ≤0.05); NE-110m single-pair arrangements 0.04–0.24 ms (budget ~1 ms); dense-synthetic full build ~30 ms for 63,500 nodes / 125,500 edges.

## Not in this PR (phase 2/3, per the design doc)

Half-edge graph, `OverlayLabel`/labeller, ring builders (phase 2); the ops, `symdifference`, mixed dimensions, and the public opt-in `OverlayNG{M}` algorithm with differential validation vs GEOS and s2 (phase 3). Existing clipping behavior is completely untouched.

### DIFF
--- a/docs/plans/2026-07-16-overlayng-noding-substrate.md
+++ b/docs/plans/2026-07-16-overlayng-noding-substrate.md
@@ -1,0 +1,377 @@
+# OverlayNG port — exact-arrangement design, and the phase 1 (noding substrate) implementation spec
+
+Status: design settled 2026-07-16, after a four-agent survey of JTS `overlayng`/`noding`,
+s2geometry's boolean-operation stack, and this repo's foundation, followed by four validation
+spikes (S1–S4, verdicts in the appendix). Phase 1 is specified here at implementation depth;
+phases 2–3 are outlined so phase-1 authors know their consumer. This document is written to be
+consumed by implementation subagents: where it names an existing function, treat the name as a
+strong pointer but verify the signature against the code before use.
+
+Reference checkouts (read-only): `/Users/anshul/temp/GO_jts/jts` (JTS, see
+`modules/core/src/main/java/org/locationtech/jts/operation/overlayng/` and `…/noding/`),
+`/Users/anshul/temp/GO_jts/s2geometry`. This repo: branch `relateng` (tip ≥ `69e416484`).
+
+---
+
+## 0. The one governing decision
+
+Overlay (`intersection` / `union` / `difference` / `symdifference`) is computed as an **exact
+arrangement with symbolic nodes, rounded once at emission** — the extension of the RelateNG
+port's design D2 to constructive output.
+
+- **No constructed coordinate ever enters a decision.** Node identity, ordering, angular sort,
+  labeling, and result assembly are all decided by exact kernel predicates over input vertices
+  and symbolic crossing keys. Float64 values appear only as *filters* (certified bounds,
+  escalate on uncertainty) and at final *emission*.
+- **There is no snapping, no snap-rounding, no precision model, no retry ladder — permanently.**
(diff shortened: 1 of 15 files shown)


asinghvi17:

PR #445, opened 2026-07-17
Add a fast WKB polygon codec to GeometryOpsTestHelpers

asinghvi17:
Adds a fast, dependency-light WKB codec for polygonal geometry to **GeometryOpsTestHelpers**: `parse_wkb` and `write_wkb`. Independent of the relateng/overlayng stack — targets `main` directly.

## Why

The upcoming differential-validation harnesses exchange geometry with C libraries as WKB (s2geography's C API speaks `geoarrow.wkb`; LibGEOS speaks WKB natively). WKB is the right interchange for differential testing — binary doubles round-trip bit-exactly, with no text-precision hazards — but the parse side is hot and shouldn't pay LibGEOS object-tree costs, and the helpers shouldn't grow a full WKB-stack dependency for two geometry types.

## What's here

`GeometryOpsTestHelpers/src/wkb.jl` (~150 SLOC):

- `parse_wkb(bytes)` → `GI.Polygon`/`GI.MultiPolygon` with `Vector{Tuple{Float64,Float64}}` rings — the exact `GO.tuples` representation (built through inner constructors, so results compare `==` coordinate-exact and empty geometries construct cleanly). Single pass with pointer loads (no `IOBuffer`, no per-value `read`); both byte orders, including **mixed endianness across multipolygon elements** (each element carries its own header, per spec); EWKB SRID accepted and skipped; Z/M (ISO thousands-digit or EWKB flag bits) rejected with a clear error; every count validated against the remaining buffer **before** allocation, so truncated or hostile buffers (e.g. a `0xFFFFFFFF` ring count) throw a clean `ArgumentError` instead of segfaulting or OOM-ing.
- `write_wkb(geom)` → little-endian ISO WKB from any GeoInterface `PolygonTrait`/`MultiPolygonTrait` geometry: exact size computed up front, pointer stores, doubles bit-for-bit. Byte-identical to LibGEOS's writer output for LE ISO polygons/multipolygons (asserted in tests).

Scope is deliberately polygonal: types 3 and 6 only; anything else throws an `ArgumentError` naming the type found.

## Validation

`test/wkb.jl` (1,640 assertions): coordinate-exact round-trips in both directions against LibGEOS (unit square, multi-hole polygons, multipolygons, empties, degenerate rings, float-stress values checked via `reinterpret(UInt64, …)` bit equality); 1,200 seeded-fuzz checks (400 random geometries × cross-checks); big-endian and mixed-endianness fixtures; EWKB SRID/Z/M fixtures; malformed-buffer cases; and a data-availability-gated Natural Earth round-trip of all 177 110m countries.

## Why not just depend on WellKnownGeometry.jl?

That was measured rather than assumed (WKG v0.2.6, Julia 1.12.6, canonical input bytes produced by the LibGEOS C writer). The comparison is like-for-like: both sides start at a byte buffer and end at the identical concrete tuple representation, and WKG is credited with its own README-recommended `GI.coordinates` traversal rather than the `GO.tuples` route (whose indexed access hits an O(i²) accessor path in WKG and would look worse still — 2.92 ms on NE 110m).

| corpus | operation | this codec | WellKnownGeometry | LibGEOS |
|---|---|---|---|---|
| NE 110m (10,654 v) | parse | **16.9 µs** (1.6 ns/v) | 511 µs (48 ns/v) | 995 µs (93 ns/v) |
| NE 10m Canada (68,193 v) | parse | **35.8 µs** (0.5 ns/v) | 3.05 ms (45 ns/v) | 5.58 ms (82 ns/v) |
| unit square | parse | **21.6 ns** | 658 ns | 668 ns |
| NE 110m | write | **14.3 µs** | 267 µs | 241 µs |
| NE 10m Canada | write | **28.3 µs** (3 allocs) | 14.9 ms (866 MB churn) | 1.01 ms |

Parse is 30–85× faster, write 11–525×; allocations are 6× smaller and 20× fewer on parse. A control run — `GO.tuples` on an *already-materialized* geometry — shows the traversal floor is only 0.1–2% of WKG's time in every regime, so the gap is decoder cost, not GeometryOps overhead. (Corollary: `parse_wkb` from bytes is 2.3× faster than `GO.tuples` on the equivalent in-memory geometry, so routing through any non-native reader would cost more than this entire decode.) There is also a capability gap: WKG rejects big-endian WKB outright, which this codec supports along with mixed-endian multipolygons.

Cross-validation against WKG at the same time: all 435 Natural Earth countries (110m + 10m, 548,471 vertices) parse coordinate-identically under both, and this writer's output is **byte-for-byte identical to both WKG's and GEOS's** — plus 2,000 fuzz polygons, special floats (−0.0, NaN payloads, denormals) bit-exact, and all six malformed-input classes throwing cleanly.

## Test plan

- [x] CI green on the tip `33b09d525` (full package suite, 5 Julia matrices) — run 29553267374, all jobs

### DIFF
--- a/GeometryOpsTestHelpers/src/wkb.jl
+++ b/GeometryOpsTestHelpers/src/wkb.jl
@@ -1,0 +1,264 @@
+# # Fast WKB codec for polygonal geometry
+export parse_wkb, write_wkb
+
+#=
+## What is this?
+
+A dependency-light, single-pass reader and writer for **Well-Known Binary (WKB)**
+restricted to polygonal geometry (`Polygon` and `MultiPolygon`).  It exists to move
+geometry in and out of C libraries (s2geography's C API, LibGEOS) as `geoarrow.wkb`
+(standard ISO WKB) during differential-validation testing, without pulling in a full
+WKB stack.
+
+* [`parse_wkb`](@ref) reads a byte buffer and returns a `GI.Polygon` or `GI.MultiPolygon`
+  whose rings are `Vector{Tuple{Float64,Float64}}` — the same nested-tuple representation
+  `GO.tuples` produces, so results compare `==` coordinate-exact against it.
+* [`write_wkb`](@ref) serializes any GeoInterface `PolygonTrait`/`MultiPolygonTrait`
+  geometry to little-endian ISO WKB.
+
+Parsing is the hot path (results come back from C as WKB), so the reader is a single
+pass over the buffer using pointer loads — no `IOBuffer`, no per-value `read`, no
+intermediate geometry objects.  Every count is validated against the remaining buffer
+length *before* any allocation, so a truncated or hostile buffer throws a clean error
+instead of segfaulting or OOMing on a giant preallocation.
+
+## WKB layout (ISO)
+
+Each geometry is: a **byte-order** byte (`0x00` big-endian/XDR, `0x01` little-endian/NDR),
+a `UInt32` **type code**, then a type-specific payload.  A `MultiPolygon` payload is a
+`UInt32` count followed by full sub-geometries — **each carrying its own byte-order byte
+and type code**, so mixed endianness within one buffer is legal and handled.
+
+Dimensionality is detected two ways and rejected (2D only): ISO codes (`type + 1000` Z,
+`+ 2000` M, `+ 3000` ZM) and EWKB high-bit flags (`0x80000000` Z, `0x40000000` M).  The
+EWKB SRID flag (`0x20000000`) is accepted and its 4-byte SRID skipped.
+=#
+
+import GeoInterface as GI
+
+# Concrete geometry types.  These match exactly what `GO.tuples` returns for 2D
+# polygonal geometry, so `parse_wkb(...) == GO.tuples(geom)` holds structurally.
+# Building them through the inner constructors (rather than the public `GI.Polygon`
+# etc.) skips validation and, crucially, works for empty rings/polygons/multipolygons
+# (the public constructors call `first(geom)` and throw on empty input).
+const _WKBRing  = Vector{Tuple{Float64,Float64}}
+const _WKBLinRing = GI.LinearRing{false,false,_WKBRing,Nothing,Nothing}
+const _WKBRings = Vector{_WKBLinRing}
+const _WKBPoly  = GI.Polygon{false,false,_WKBRings,Nothing,Nothing}
+const _WKBPolys = Vector{_WKBPoly}
+const _WKBMPoly = GI.MultiPolygon{false,false,_WKBPolys,Nothing,Nothing}
+
+@inline _wkb_linearring(coords::_WKBRing) = _WKBLinRing(coords, nothing, nothing)
+@inline _wkb_polygon(rings::_WKBRings) = _WKBPoly(rings, nothing, nothing)
+@inline _wkb_multipolygon(polys::_WKBPolys) = _WKBMPoly(polys, nothing, nothing)
+
+# WKB integer type codes -> human names, for error messages.
+function _wkb_type_name(code::Integer)
+    code == 1 ? "Point" :
+    code == 2 ? "LineString" :
+    code == 3 ? "Polygon" :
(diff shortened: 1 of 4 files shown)


GeometryOpsTestHelpers/src/wkb.jl
```diff
@@ -0,0 +1,300 @@
...
+    end
+    return _WKBPolygon(rings, nothing, nothing), pos
+end
+
+# Parse a MultiPolygon body (header already consumed); return `(multipolygon, pos)`.
+@inline function _read_multipolygon_body(bytes, pos::Int, n::Int, le::Bool)
+    ngeoms32, pos = _read_u32(bytes, pos, n, le)
+    ngeoms = Int(ngeoms32)
+    # Each sub-polygon is at least order(1) + type(4) + nrings(4) = 9 bytes.
+    Int64(pos) + Int64(ngeoms) * 9 <= n || _wkb_truncated(ngeoms * 9, n - pos)
```
asinghvi17: Why force convert to Int64 here - what value does it add?

GeometryOpsTestHelpers/src/wkb.jl
```diff
@@ -0,0 +1,300 @@
...
+Base.showerror(io::IO, e::WKBMultiPolygonElementError) = print(io,
+    "WKBMultiPolygonElementError: a MultiPolygon element must be a Polygon (3), but found type " *
+    "$(e.code) ($(_wkb_type_name(e.code)))")
+
+# Cold throw paths, kept out of the inlined readers.
+@noinline _wkb_truncated(need, have) = throw(WKBTruncatedError(need, have))
+@noinline _wkb_bad_order(b) = throw(WKBByteOrderError(b))
+@noinline _wkb_zm(rawtype) = throw(WKBDimensionError(rawtype))
+@noinline _wkb_unsupported(code) = throw(WKBUnsupportedTypeError(code))
+@noinline _wkb_mp_child(code) = throw(WKBMultiPolygonElementError(code))
```
asinghvi17: Just call throw inline, re benchmark if you are afraid of performance loss

GeometryOpsTestHelpers/src/wkb.jl
```diff
@@ -0,0 +1,300 @@
...
+@inline function _read_polygon_body(bytes, pos::Int, n::Int, le::Bool)
+    nrings32, pos = _read_u32(bytes, pos, n, le)
+    nrings = Int(nrings32)
+    # Each ring is at least its own 4-byte point count; bound the allocation first.
+    Int64(pos) + Int64(nrings) * 4 <= n || _wkb_truncated(nrings * 4, n - pos)
+    rings = Vector{_WKBLinearRing}(undef, nrings)
+    @inbounds for r in 1:nrings
+        npts32, pos = _read_u32(bytes, pos, n, le)
+        npts = Int(npts32)
+        need = Int64(npts) * 16
```
asinghvi17: again, double converting seems weird

GeometryOpsTestHelpers/src/wkb.jl
```diff
@@ -0,0 +1,300 @@
...
+end
+Base.showerror(io::IO, e::WKBUnsupportedTypeError) = print(io,
+    "WKBUnsupportedTypeError: type $(e.code) ($(_wkb_type_name(e.code))) is not Polygon (3) or MultiPolygon (6)")
+
+# A MultiPolygon holds a sub-geometry that is not a Polygon.
+struct WKBMultiPolygonElementError <: WKBParseError
+    code::Int
+end
+Base.showerror(io::IO, e::WKBMultiPolygonElementError) = print(io,
+    "WKBMultiPolygonElementError: a MultiPolygon element must be a Polygon (3), but found type " *
```
asinghvi17:
```suggestion
    "WKBMultiPolygonElementError: a MultiPolygon element must be a Polygon (code 3), but found code " *
```

asinghvi17:
To clarify, this is only going to GOTestHelpers so not production code.  When this is ready it can replace the WellKnownGeometry codebase or augment it.

PR #451, opened 2026-08-02
Implement spatial join shorthand

evetion:
As suggested by @rafaqz in #113, but without introducing new GO specific join methods.


asinghvi17 (approved):
Looks good to me, thanks!

PR #436, opened 2026-07-11
Add manifold-aware construction to `RTree` and `NaturalIndex`

asinghvi17:
Stacked on #434 (the base branch is `spherical-arc-extent`); will be rebased and retargeted to `main` once that merges.

## What this does

`RTree` and `NaturalIndex` already accept a vector of extents, so making them manifold-aware is one delegating method each:

```julia
RTree(m::Manifold, algorithm, data; nodecapacity = 16)
NaturalIndex(m::Manifold, geoms; nodecapacity = 32)
```

Each maps its elements through `Extents.extent(m, x)` (from #434) and calls the existing extent-vector constructor. On `Spherical()` the leaf boxes are the 3D Cartesian boxes of the elements as regions on the unit sphere — covering the arc bulge and enclosed poles that the default `GI.extent` vertex boxes miss. Leaf indices still map back to positions in the input collection.

This closes the loop with #432's `Extents.intersects(::SphericalCap, ::Extent{(:X, :Y, :Z)})`:

```julia
band = [GI.Polygon([[(lon, 60.0), (lon + 30.0, 60.0), (lon + 30.0, 80.0), (lon, 80.0), (lon, 60.0)]])
        for lon in 0.0:30.0:330.0]
cap  = GI.Polygon([[(lon, 80.0) for lon in 0.0:30.0:360.0]])   # a cell around the north pole
tree = RTree(GO.Spherical(), HPR(), vcat(band, [cap]))

polecap = SphericalCap(UnitSphericalPoint(0.0, 0.0, 1.0), 0.05)
STI.query(tree, Base.Fix1(Extents.intersects, polecap))        # == [13]: the cap, and only the cap
```

The query side is deliberately **not** manifold-overloaded: a query has to be in the space the tree was built in, and a hypothetical `query(tree, geographic_geom)` convenience would go through `GI.extent` and silently compare a lon/lat box against XYZ boxes over the shared `X`/`Y` keys. Convert explicitly — `GO.extent(GO.Spherical(), geom)` — as the docs now spell out.

## Docs

- New **"Extents and spatial indexing"** section in the manifolds explanation: why vertex boxes under-cover on the sphere, what `Extents.extent(m::Manifold, geom)` returns, the manifold-aware tree constructors, and the same-manifold caveat — with executable `@example` blocks.
[...]

PR #452, opened 2026-08-02
Add spatialtree to get existing tree, or create one

evetion:
for a given collection. Incomplete AI draft with some touches, to go together with a PR to GeoDataFrames (always creating a FlexibleRTree on reading). I would like to use that tree, but at the moment GO always creates it.

Probably `spatialtree` should get expanded to support creation of different trees? We probably need a short design session.

This also adds the ability to add a custom indices, because geometries in a vector can be missing/nothing.

### DIFF
--- a/ext/GeometryOpsFlexiJoinsExt/GeometryOpsFlexiJoinsExt.jl
+++ b/ext/GeometryOpsFlexiJoinsExt/GeometryOpsFlexiJoinsExt.jl
@@ -52,7 +52,7 @@ function FlexiJoins.findmatchix(::FlexiJoins.Mode.Tree, cond::FlexiJoins.ByPred{
     # We extract the relevant columns using cond.Lf and cond.Rf.
     tree === nothing && return Int[]
     left_geom = cond.Lf(a)
-    (left_geom === nothing || ismissing(left_geom) || isnothing(GI.extent(left_geom))) && return Int[]
+    (isnothing(left_geom) || ismissing(left_geom) || GI.isempty(left_geom)) && return Int[]
     idxs = query(tree, left_geom)
     intersecting_idxs = filter!(idxs) do idx
         cond.pred(cond.Lf(a), cond.Rf(B[idx]))
--- a/ext/GeometryOpsFlexiJoinsExt/GeometryOpsFlexiJoinsExt.jl
+++ b/ext/GeometryOpsFlexiJoinsExt/GeometryOpsFlexiJoinsExt.jl
@@ -13,7 +13,9 @@ using GeometryOps
 using FlexiJoins
 
 import GeometryOps as GO, GeoInterface as GI
-using SortTileRecursiveTree, Tables
+import GeometryOps.SpatialTreeInterface: spatialtree
+using GeometryOps.SpatialTreeInterface: query, isspatialtree
+using Tables
 
 
 # This module defines the FlexiJoins APIs for GeometryOps' boolean comparison functions, taken from DE-9IM.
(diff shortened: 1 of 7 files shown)


asinghvi17:
We should definitely have a design session.  Manifold awareness etc will be crucial to this working but I think we can do it

src/utils/FlexibleRTrees/interface.jl
```diff
@@ -85,3 +85,17 @@ function query(tree::RTree, geom)
     isnothing(ext) && throw(ArgumentError("no extent found on $(typeof(geom))"))
     return query(tree, ext)
 end
+
+function spatialtree(geometries)
+    valid = nonmissingtype(eltype(geometries))[]
+    positions = Int[]
+    for (i, geom) in enumerate(geometries)
+        if geom === nothing || ismissing(geom) || isnothing(GI.extent(geom))
```
asinghvi17:
```suggestion
        if geom === nothing || ismissing(geom)
```

src/utils/FlexibleRTrees/interface.jl
```diff
@@ -85,3 +85,17 @@ function query(tree::RTree, geom)
     isnothing(ext) && throw(ArgumentError("no extent found on $(typeof(geom))"))
     return query(tree, ext)
 end
+
+function spatialtree(geometries)
```
asinghvi17: Let's thread a manifold through here, let the default be planar and then inject the manifold into the first argument of the tree constructor

ext/GeometryOpsFlexiJoinsExt/GeometryOpsFlexiJoinsExt.jl
```diff
@@ -35,12 +37,23 @@ FlexiJoins.supports_mode(::FlexiJoins.Mode.Tree, ::FlexiJoins.ByPred{F}, datas)
...
+
+FlexiJoins.prepare_for_join(::FlexiJoins.Mode.Tree, X, cond::FlexiJoins.ByPred{<: GO_DE9IM_FUNCS}) = (X, spatialtree(X, cond.Rf))
+function FlexiJoins.findmatchix(::FlexiJoins.Mode.Tree, cond::FlexiJoins.ByPred{F}, ix_a, a, (B, tree)::Tuple, multi::typeof(identity)) where F<:GO_DE9IM_FUNCS
     # Implementation note:
     # here, `a` is a row, and `b` is the full table.
     # We extract the relevant columns using cond.Lf and cond.Rf.
-    idxs = SortTileRecursiveTree.query(tree, cond.Lf(a))
+    tree === nothing && return Int[]
+    left_geom = cond.Lf(a)
+    (left_geom === nothing || ismissing(left_geom) || isnothing(GI.extent(left_geom))) && return Int[]
```
asinghvi17:
```suggestion
    (isnothing(left_geom) || ismissing(left_geom) || GI.isempty(left_geom)) && return Int[]
```

src/utils/FlexibleRTrees/interface.jl
evetion: There's no such thing in FlexibleRTrees or SpatialTreeInterface right? I'd say leave it for later improvement.

asinghvi17: Ah right, that's in https://github.com/JuliaGeo/GeometryOps.jl/pull/436 and will be merged in the next couple days to main.

Let's take it as a first arg here but do nothing with it for now?  Before releasing I will hook that up.

asinghvi17:
Hmm tests seem to be failing for some reason

src/utils/FlexibleRTrees/interface.jl
evetion: If we have a default it should be the last argument no?

src/utils/FlexibleRTrees/interface.jl
```diff
@@ -85,3 +85,17 @@ function query(tree::RTree, geom)
...
+    for (i, geom) in enumerate(geometries)
+        if (isnothing(geom) || ismissing(geom) || GI.isempty(geom))
+            continue
+        end
+        push!(valid, geom)
+        push!(positions, i)
+    end
+    isempty(valid) && return nothing
+    return RTree(STR(), valid; indices=positions)
+end
```
asinghvi17:
```suggestion
spatialtree(geometries) = spatialtree(Planar(), geometries)
function spatialtree(manifold::GOCore.Manifold, geometries)
    valid = nonmissingtype(eltype(geometries))[]
    positions = Int[]
    for (i, geom) in enumerate(geometries)
        if (isnothing(geom) || ismissing(geom) || GI.isempty(geom))
            continue
        end
        push!(valid, geom)
        push!(positions, i)
    end
    isempty(valid) && return nothing
    return RTree(manifold, STR(), valid; indices=positions)
end
```

PR #453, opened 2026-08-03
OverlayNG phase 3: the four ops, public `OverlayNG{M}` algorithm, and differential validation

asinghvi17:
Phase 3 of the OverlayNG port — the four ops reach a public surface, and the engine gets the validation it needs to be trusted. Stacked on #444; review that first.

Design: `docs/plans/2026-07-16-overlayng-noding-substrate.md` §4.

## Public surface

`OverlayNG{M} <: Algorithm{M}` is **opt-in**; Foster–Hormann remains the default for `intersection`/`union`/`difference`, and no existing default behaviour changes.

```julia
intersection(OverlayNG(), a, b)             # planar
union(OverlayNG(Spherical()), a, b)         # spherical
symdifference(a, b)                         # new op, no Foster–Hormann equivalent
```

Point inputs and mixed-dimension results are handled per `OverlayUtil`'s rules.

## Validation

- **JTS XML conformance** — 768/768, plus the `robust/overlay` corpus run as a ratchet.
- **Differential fuzz vs LibGEOS** — 1600/1600 across 8 generators, 0 divergent. GEOS ≥3.9 *is* OverlayNG, so this is same-algorithm-to-same-algorithm. A deep sweep of 80,000 ops is clean.
- **Real-data identities** — Natural Earth 110m/10m area conservation, worst residual ~6e-15.
- **Spherical differential vs s2geography** — new, and the only oracle that can validate geodesic edges at all, since GEOS is planar. 204 ops, median area agreement 3.3e-15. Also a predicate differential against `RelateNG(; manifold = Spherical())`: 120 comparisons, exact agreement.

`S2Geography_jll` is a new test-only dependency; all five CI matrices resolve a binary, and the suite skips gracefully when it is absent.

## Fixes surfaced by the validation

- **Result assembly was deciding shell-vs-hole on rounded output coordinates**, which design §0 reserves for exact kernel predicates. Taking orientation from exact node coordinates eliminates all 56 `OverlayTopologyError`s in the robust corpus, raises GEOS area agreement from 186 to 223 of 260 op-runs, and cures the corpus's worst area residual (4.44e-2 → 1.33e-11) to GEOS-identical.
- **Areal inputs were never self-noded.** A valid polygon can self-touch at a point interior to another ring's segment. Noding each input's vertices against its own segments is provably sufficient here — verified identical to a full all-pairs pass on 5,445 corpus pairs, with the single divergence occurring on invalid input.
- **Face-walk hygiene** — opt-in dangle and cut-edge removal, a guard against reusing a consumed graph (which previously returned silently wrong rings), and one shared ring-selection path instead of two.

## Known limits

50 robust-corpus cases remain ledgered in `test/external/jts/overlay_skiplist.jl`, each with a diagnosis. They are correct-area results whose rounded linework GEOS calls `Self-intersection` — sub-ULP slivers the exact arrangement keeps and Float64 output cannot represent. Reaching GEOS's answer on them requires snapping, which this design deliberately does not do.


asinghvi17:

PR #460, opened 2026-08-12
Let a prepared geometry stand in for the A geometry in `RelateNG`

asinghvi17:
## Motivation

`RelateNG`'s named predicates — `GO.covers(alg, a, b)` and the other nine — rebuild the A side on every call. A user who had already `prepare`d the A geometry had no way to hand it to them: `covers(RelateNG(...), prep, poly)` was a `MethodError`, so the only prepared spelling was

```julia
prep = GO.prepare(GO.RelateNG(GO.Spherical()), texas)
for poly in polys
    GO.relate_predicate(prep, GO.pred_covers(), poly)   # the only prepared spelling
end
```

and the natural thing

```julia
for poly in polys
    GO.covers(GO.RelateNG(GO.Spherical()), texas, poly)  # rebuilds texas on every call
end
```

silently rebuilt the whole prepared structure — point locators, edge index, spherical ring validation, on the order of 100 ms — on *every* call, defeating `prepare` entirely.

## After

```julia
prep = GO.prepare(GO.RelateNG(GO.Spherical()), texas)
for poly in polys
    GO.covers(GO.RelateNG(GO.Spherical()), prep, poly)
end
```

A `PreparedRelate` now stands in for the A geometry at every `RelateNG` entry point, so a call site already written against `covers(alg, a, b)` can swap a prepared `a` in without any other change:

```julia
GO.relate(alg, prep, b)
GO.relate(alg, prep, b, "T*T***T**")
GO.relate_predicate(alg, GO.pred_covers(), prep, b)
GO.covers(alg, prep, b)                                  # …and the other nine
```

## What this adds

Two methods:

1. **`prepare(alg, p::PreparedRelate)` is idempotent** — it returns `p` itself when `alg == p.alg`, and otherwise prepares afresh from `p`'s input geometry. Every `RelateNG` setting is baked into either the prepared `RelateGeometry` and its locators (`manifold`, `boundary_rule`, `exact`) or the index built over the segment strings (`accelerator`), so a mismatched algorithm cannot reuse the cache.

2. **`relate_predicate(alg, pred, p::PreparedRelate, b)`** — the single dispatch point. `relate`, the pattern form and all ten named predicates already forward to `relate_predicate`, so they need no methods of their own.

## One structural change

`PreparedRelate` gained an `input` field holding the geometry exactly as passed to `prepare`. The rebuild path needs it: `geom_a.geom` is an extent-cached wrapper tree whose cached extents are *that* manifold's interaction bounds (3-D on `Spherical`), and re-preparing it under a different manifold silently reuses those as planar boxes. That produced wrong answers — `covers` returning `false` where it should return `true` — until the cross-manifold tests caught it. Holding the input costs nothing: the wrapper tree already shares the input's linework.

## Tests

Added to `test/methods/relateng/relate_ng.jl`:

- all ten named predicates as `f(alg, prep, b)` against `f(alg, a, b)` over ten B geometries covering each predicate's true and false cases, on both `Planar()` and `Spherical()`, plus the `relate`, pattern and `relate_predicate` forms;
- `prepare` idempotence (`===` identity), and rebuild on a change to each of `manifold`, `boundary_rule`, `exact` and `accelerator`;
- cross-algorithm queries returning the answer the *passed* algorithm defines, including a `Mod2Boundary` vs `EndpointBoundary` case where the two rules genuinely disagree (a valence-2 node of a `MultiLineString`).

No existing behaviour changed, no new names exported.

### DIFF
--- a/src/methods/geom_relations/relateng/relate_ng.jl
+++ b/src/methods/geom_relations/relateng/relate_ng.jl
@@ -236,6 +236,12 @@ dispatching to the old engines (design D4).
 DE-9IM `T*F**FFF*` sense), which can differ from the structural equality
 the two-argument `GO.equals` implements only in exotic cases (both
 treat rotated/reversed rings and repeated points as equal).
+
+These rebuild the A side on every call. When one geometry is queried many
+times, [`prepare`](@ref) it: a `PreparedRelate` may be passed as `g1` to
+any of these (it flows through `relate_predicate`, which reuses it rather
+than rebuilding), or the two-argument prepared forms `GO.covers(prep, b)`
+may be used directly. See the prepared-mode section below.
 ==========================================================================#
 "This functionality is experimental and may change at any time."
 intersects(alg::RelateNG, g1, g2) = relate_predicate(alg, pred_intersects(), g1, g2)
@@ -623,7 +629,7 @@ branches.
 ==========================================================================#
 
 """
-    PreparedRelate{ALG, RG, SS, T}
+    PreparedRelate{ALG, G, RG, SS, T}
 
 This functionality is experimental and may change at any time.
 
@@ -632,6 +638,11 @@ topological relationships against a single geometry `a` (the "prepared
 mode" of JTS `RelateNG.prepare`). Holds:
 
 - `alg`: the [`RelateNG`](@ref) algorithm configuration,
+- `input`: the A geometry exactly as passed to [`prepare`](@ref). Held
+  because `geom_a` below is *manifold-specific* — its cached extents are
+  the interaction bounds of `alg`'s manifold (3D on `Spherical`), so it
+  cannot be re-prepared under a different one; the raw input can. Costs
+  nothing: the wrapper tree already shares the input's linework,
 - `geom_a`: the A-side [`RelateGeometry`](@ref), constructed with
   `is_prepared = true` and with its lazy locator/unique-points caches
   forced,
@@ -650,9 +661,10 @@ Construct with [`prepare`](@ref); evaluate with [`relate`](@ref) /
     `RelateGeometry` (edge re-extraction, element-id counter). Use one
     `PreparedRelate` per thread.
 """
-struct PreparedRelate{ALG <: RelateNG, RG <: RelateGeometry,
+struct PreparedRelate{ALG <: RelateNG, G, RG <: RelateGeometry,
         SS <: AbstractVector{<:RelateSegmentString}, T <: Union{Nothing, RTree}}
     alg::ALG
+    input::G
     geom_a::RG
     segs_a::SS
     edge_tree::T
@@ -733,7 +745,7 @@ function prepare(alg::RelateNG, a;
     #-- force the lazy caches that repeated evaluations reuse
     _get_locator(geom_a)
     get_dimension_real(geom_a) == DIM_P && get_unique_points(geom_a)
-    return PreparedRelate(alg, geom_a, segs_a, edge_tree)
+    return PreparedRelate(alg, a, geom_a, segs_a, edge_tree)
 end
 
 # The manifold-dependent `validate` default of `prepare` (see its docstring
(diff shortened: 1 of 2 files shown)


src/methods/geom_relations/relateng/relate_ng.jl
```diff
@@ -743,6 +755,86 @@ end
...
+`Spherical(radius = 1)` and `Spherical(radius = 1.0)` are not `===`. Those
+are compared parameter-by-parameter with numeric `==` instead; anything
+else (including a manifold type this file does not know about) falls back
+to the strict comparison, which errs toward reporting a difference.
+=#
+_manifolds_agree(m1::Manifold, m2::Manifold) = m1 == m2
+_manifolds_agree(m1::Spherical, m2::Spherical) =
+    m1.radius == m2.radius && m1.oriented == m2.oriented
+_manifolds_agree(m1::Geodesic, m2::Geodesic) =
+    m1.semimajor_axis == m2.semimajor_axis && m1.inv_flattening == m2.inv_flattening
```
asinghvi17: sounds like these can all be just `==` and we can drop this function

src/methods/geom_relations/relateng/relate_ng.jl
```diff
@@ -918,6 +1010,150 @@ satisfies the predicate. Port of the instance method
...
+accepts a `PreparedRelate` as its A geometry — `relate(alg, p, b)`,
+`relate(alg, p, b, im_pattern)` and the named predicates
+`covers(alg, p, b)` etc. all forward here — so all of them reuse the
+prepared structures rather than rebuilding A on every call.
+
+!!! note
+    Predicates are mutable accumulators — pass a freshly constructed one
+    (e.g. `pred_intersects()`) per evaluation.
+"""
+relate_predicate(alg::RelateNG, predicate::TopologyPredicate, p::PreparedRelate, b) =
```
asinghvi17, line 946: it should not exist this function

src/methods/geom_relations/relateng/relate_ng.jl
```diff
@@ -918,6 +1010,150 @@ satisfies the predicate. Port of the instance method
...
+"""
+
+"""
+    intersects(p::PreparedRelate, b)::Bool
+
+Tests whether the prepared geometry intersects `b`.
+
+$(_PREPARED_PREDICATE_DOC)
+"""
+intersects(p::PreparedRelate, b) = relate_predicate(p, pred_intersects(), b)
```
asinghvi17: none of this should exist, it should just work with intersects(RelateNG(...), geom1, geom2)

src/methods/geom_relations/relateng/relate_ng.jl
asinghvi17: Dropped. `prepare` now just compares algorithms with `==`:

```julia
prepare(alg::RelateNG, p::PreparedRelate; validate = ...) =
    alg == p.alg ? p : prepare(alg, p.input; validate)
```

All the `RelateNG` fields and the manifolds are immutable, so `==` falls back to `===`, which is exact for identically-constructed algorithms. The only case it loses is `Spherical(radius = 1)` vs `Spherical(radius = 1.0)`, which now rebuilds instead of reusing — a conservative miss, not a wrong answer.

The `_prepared_geometry_settings_agree` split went with it, so an `accelerator` difference rebuilds too rather than swapping just the index.

src/methods/geom_relations/relateng/relate_ng.jl
asinghvi17: Kept, but down to two lines and a comment — this is the dispatch that makes `intersects(RelateNG(...), prep, geom2)` work at all. The unprepared `relate_predicate(alg, pred, a, b)` builds a `RelateGeometry` from `a`, so a prepared A side needs *somewhere* to branch, and putting it here is the cheapest place: `relate`, the pattern form and all ten named predicates already forward to `relate_predicate`, so one method covers every entry point.

```julia
relate_predicate(alg::RelateNG, predicate::TopologyPredicate, p::PreparedRelate, b) =
    relate_predicate(prepare(alg, p), predicate, b)
```

The only alternative I see is an `a isa PreparedRelate` branch inside the unprepared method, which is the same thing spelled worse. Say the word if you would rather have that.

src/methods/geom_relations/relateng/relate_ng.jl
asinghvi17: All ten removed, along with the shared docstring constant. `intersects(RelateNG(...), prep, geom2)` is the only prepared named-predicate spelling now; `relate(prep, b)` and `relate_predicate(prep, pred, b)` predate this PR and are untouched.

src/methods/geom_relations/relateng/relate_ng.jl
```diff
@@ -918,6 +962,14 @@ satisfies the predicate. Port of the instance method
 relate_predicate(p::PreparedRelate, predicate::TopologyPredicate, b) =
     evaluate!(p.alg, p.geom_a, b, predicate, p)
 
+# A `PreparedRelate` stands in for the A geometry at every `RelateNG` entry
+# point: `relate(alg, p, b)`, `relate(alg, p, b, im_pattern)` and the named
+# predicates `covers(alg, p, b)` etc. all forward here, and `prepare` returns
+# `p` itself under the algorithm it was prepared with, so the cached A side is
+# reused instead of rebuilt.
```
asinghvi17: Compact this comment

src/methods/geom_relations/relateng/relate_ng.jl
```diff
@@ -632,6 +637,11 @@ topological relationships against a single geometry `a` (the "prepared
 mode" of JTS `RelateNG.prepare`). Holds:
 
 - `alg`: the [`RelateNG`](@ref) algorithm configuration,
+- `input`: the A geometry exactly as passed to [`prepare`](@ref). Held
+  because `geom_a` below is *manifold-specific* — its cached extents are
+  the interaction bounds of `alg`'s manifold (3D on `Spherical`), so it
+  cannot be re-prepared under a different one; the raw input can. Costs
+  nothing: the wrapper tree already shares the input's linework,
```
asinghvi17: Don't tell a story here

src/methods/geom_relations/relateng/relate_ng.jl
```diff
@@ -743,6 +754,39 @@ end
...
+index built over the segment strings.
+
+!!! warning
+    That rebuild is paid on *every* call: preparing under one algorithm and
+    querying under another — `prep = prepare(RelateNG(Spherical()), a)`
+    followed by `covers(RelateNG(), prep, b)` in a loop — is exactly what
+    `prepare` exists to avoid. Prepare under the algorithm you query with.
+
+`validate` applies only to the rebuild; a reused prepared geometry keeps
+whatever validation state it was built with.
```
asinghvi17: way too long

src/methods/geom_relations/relateng/relate_ng.jl
asinghvi17: Two lines:

```julia
# The prepared A side at the algorithm entry points: `relate` and the named
# predicates all forward here, so this is the only method they need.
```

src/methods/geom_relations/relateng/relate_ng.jl
asinghvi17: Cut to the length of the bullets around it:

```julia
- `input`: the A geometry as passed to [`prepare`](@ref), which re-prepares
  from it under a different `alg` (`geom_a` below caches `alg`'s manifold's
  extents, so it cannot serve another),
```

src/methods/geom_relations/relateng/relate_ng.jl
asinghvi17: Down to one paragraph — the three-case breakdown and the warning admonition are gone:

```julia
Returns `p` under the algorithm it was prepared with, so a `PreparedRelate`
can be passed as the A geometry to any `RelateNG` entry point. Under any
other `alg` — every setting is baked into the prepared structures — it is
prepared afresh from `p.input`, at the full cost of [`prepare`](@ref) on
every call, so prepare under the algorithm you query with. `validate`
applies only to that rebuild.
```

Also trimmed the paragraph I had added to the named-predicate section header down to one sentence, same reason.

PR #462, opened 2026-08-13
Carry node extents through `dual_depth_first_search`

asinghvi17:
## The problem

`dual_depth_first_search` re-derives a node's extent every time it looks at it.

#435 hoisted the calls that were loop-invariant *within a single visit*, but the recursion itself still throws the extents away. Every recursive call recomputes `node_extent` for both nodes it was just handed — even though the caller had already computed both in order to evaluate the predicate that authorised the descent. In the one-leaf branches this means the pinned node's extent is rebuilt at every level of the descent, and in the both-internal branch `node_extent(child2)` sits inside the `child1` loop, so it is recomputed once per child of node 1.

For a tree that stores an extent per node this is a field load and nobody notices. For a tree that *derives* it — a DGGS cursor whose extent is a spherical cap fitted over cell geometry, a curvilinear grid that walks the perimeter of its index rectangle — it is the dominant cost of the whole traversal.

Separately, in the both-leaves branch `child_indices_extents(node2)` is a function call in the inner `for` header, so it is re-evaluated once per cell of node 1. The interface docs encourage implementations to materialize a vector there, which makes that a per-entry allocation as well as a recomputation.

## The change

**Carry the extents through the recursion.** A new internal 6-argument form takes `(node1, extent1, node2, extent2)`; the public 4-argument signature is unchanged and just seeds them. The extents were computed anyway to evaluate the predicate, so this is free, and it removes one `node_extent` call per recursive step — `1/(1 + fanout)` of all such calls in a one-sided descent.

**Bind `child_indices_extents(node2)` before the outer loop** in the both-leaves branch.

**Add `node_extent_is_expensive(node)`, defaulting to `false`.** This is the part that needs justifying.

The remaining redundancy is the inner side of the both-internal branch, where each child's extent is re-derived once per opposing child. Materializing the children's extents removes a factor of fanout — but costs an allocation per visited internal node.

That trade is not universally good, and I measured it rather than assuming. For every tree GeometryOps actually ships and uses on its hot paths — `STRtree`, `NaturalIndex`, the `FlexibleRTrees` — `node_extent` is a stored field, so materializing buys nothing and costs real memory. `NaturalIndex`, which backs clipping and RelateNG, has `nodecapacity = 32`, so unconditional materialization would allocate a 32-element vector per visited internal node on the hottest path in the library. Measured on two 20k-edge rings, that is **~5% slower and 182 KB of garbage per call** where there is currently none.

So the materialization is opt-in. Trees whose extents are genuinely computed define the trait; everything else is untouched. The trait is constant per type, so exactly one of the two strategies is compiled into any given traversal — the default path still allocates nothing, and there is a test asserting `@allocated == 0`.

This is a deliberate revision of the note in #435 ("a full hoist that also caches the inner side's child extents was measured and rejected: the per-visit buffer doubles descent time on precomputed-extent trees"). That conclusion was right *for those trees*, and the trait is how both facts coexist.

## Correctness

Unchanged, and worth stating precisely because it is easy to state wrongly.

The traversal's soundness rests on one invariant the trees already provide: **a node's extent contains its own subtree's cells**. That is what makes a failed predicate at a node pair proof that no leaf pair beneath it can pass, which is what licenses the prune.

It does **not** require `extent(child) ⊆ extent(parent)`. That stronger property is in fact violated by some trees in this family — centroid-plus-max-radius caps are not monotone under subsetting, and on the trees that motivated this work it fails at 214/400 nodes of a DGGS tree and 384 of a curvilinear one. Nothing here depends on it.

Neither change alters which extents are compared, nor the order in which they are compared, nor the predicate. Only *how many times each extent is derived* changes. So the guarantee is exactly the one the current code has.

## Benchmarks

Reproduced here on `NaturalIndex` trees, 20k edges a side, single-threaded, against a `node_extent` that derives the extent from boundary points on access:

| tree shape | `node_extent` calls | time | allocated |
|---|---|---|---|
| nodecapacity 32, balanced | 44,664 → 2,783 | **1.42x** | 226 KB |
| nodecapacity 7, balanced | 47,688 → 11,971 | **1.59x** | 889 KB |
| nodecapacity 4, mismatched depths | 10,397 → 8,561 | **1.13x** | 0 B (carrying alone) |

and on the same trees with their real stored extents, where the trait is `false`:

| | time | allocated |
|---|---|---|
| nodecapacity 32, real extents | **1.00x** | **0 B** |

Two things worth separating out, because the reference implementation this started from conflated them:

- **Carrying extents** is free and never loses, but on *balanced* trees it saves almost nothing — the one-leaf branches, where it helps, are barely exercised. Its win shows up on mismatched depths (1.09–1.13x above).
- **Materializing** is where the large factors come from, and only when `node_extent` is expensive.

On the motivating workload — ConservativeRegridding's candidate search, IGEO7 DGGS against a lon/lat raster, 8 threads, on a build with an O(1) `node_extent` — the two together were measured at 0.776 s → 0.388 s (2.00x) at L=11 (117,649 × 264,196) and 42.46 s → 25.09 s (1.69x) at full scale (5,764,801 × 12,967,201), with end-to-end `Regridder` construction going 64.47 s → 42.12 s and the assembled sparse matrix bit-identical (nnz 23,589,936). Candidate sets were identical (808,026 pairs at L=11). **Note that reproducing those numbers now requires the DGGS cursor to define `node_extent_is_expensive`** — that is the cost of not regressing everything else.

## Trade-offs, stated plainly

- The opt-in path allocates a small vector per visited internal node. That is the one thing here that is a trade rather than a strict improvement. I considered non-allocating formulations — a fixed-size buffer needs a statically known fanout, which the interface does not provide; swapping the loop order just moves the problem to the other side; and a single-sided descent avoids it structurally but changes the traversal, which is the follow-up PR, not this one. Gating it on a trait was the only formulation I found that gets the win where it exists without charging for it where it does not.
- `node_extent_is_expensive` is new public interface surface. It has a default, so no existing implementation has to do anything.

## Tests

`test/utils/SpatialTreeInterface.jl` goes 72 → 125 passing. New coverage:

- the 4-arg and 6-arg forms agree, across `STRtree`, `NaturalIndex`, `FlatNoTree` and a varying-fanout tree, on deliberately mismatched tree depths;
- opting into `node_extent_is_expensive` changes the number of `node_extent` calls but not the answer;
- the default path allocates nothing, for both fixed and varying fanout;
- `_child_extents` folds to a concrete type from the node type alone (`@inferred`, both trait values);
- `Action(:full_return, true)` propagates out through all four branches of the recursion — both-internal, both one-leaf branches, and both-leaves — through both the 4-arg and 6-arg forms, and through an opted-in tree. The existing suite did not cover this and it is easy to break.

## Test results

Baseline established on unmodified `main` (f21f45a66) first, so nothing pre-existing is attributed here. All `dual_depth_first_search` consumers:

| suite | baseline | this PR |
|---|---|---|
| `utils/SpatialTreeInterface.jl` | 72 | **125** |
| `utils/LoopStateMachine.jl` | 19 | 19 |
| `utils/FlexibleRTrees.jl` | 1231 | 1231 |
| `methods/clipping/polygon_clipping.jl` | 6347 | 6347 |
| `methods/clipping/overlayng/noding.jl` | 3654 | 3654 |
| `methods/relateng/runtests.jl` | 147606 | 147606 |

0 failures and 0 errors throughout, before and after.

## Review follow-up (9e6ea5d)

- The lazy `(child, node_extent(child))` pair iterator was type-unstable for trees with a varying number of children: `Base.Generator{I,F}` over a non-concrete `getchild` return type is a `UnionAll`, so it boxes and every `iterate` is a dynamic dispatch. On a varying-fanout test tree the 6-arg traversal had 4 dynamic `Base.Generator(...)` call sites and allocated 45360 bytes; it is now 0 and 0 B. The traversal calls `node_extent(child)` in the loop body instead. (`enumerate` has the same failure mode, so the cached inner loop uses a plain counter.)
- `Val(node_extent_is_expensive(node))` is replaced by a branch inside `_child_extents(node::N) where {N}`, so the strategy is selected from the type rather than relying on constant propagation. Both arms are `@inferred` in the tests.
- Docstrings, the README entry and the traversal's comment blocks are cut back.

---

Draft: I would like a second opinion on the trait — both on whether `node_extent_is_expensive` is the right name and on whether gating is preferable to just eating the allocation everywhere.


src/utils/SpatialTreeInterface/dual_depth_first_search.jl
```diff
@@ -1,58 +1,99 @@
...
-        # both nodes are leaves, so we can just iterate over the indices and extents
-        for (i1, extent1) in child_indices_extents(node1)
-            for (i2, extent2) in child_indices_extents(node2)
-                if predicate(extent1, extent2)
+    return dual_depth_first_search(f, predicate, node1, node_extent(node1), node2, node_extent(node2))
+end
+
+# Children paired with their extents, for a loop that makes a single pass.  Lazy,
+# so this costs exactly one `node_extent` per child and allocates nothing.
+@inline _children_with_extents(node) = ((child, node_extent(child)) for child in getchild(node))
```
asinghvi17: nope this will be type unstable for trees with varying numbers of children.

src/utils/SpatialTreeInterface/dual_depth_first_search.jl
```diff
@@ -1,58 +1,99 @@
...
+# loops here used to do inline.  For a tree that stores an extent per node that
+# is a field load and re-deriving is free, so we keep the generator and allocate
+# nothing.  For a tree that computes its extents, `node_extent_is_expensive`
+# opts into materializing them once per node pair instead - trading a small
+# vector for a factor-of-fanout reduction in `node_extent` calls.
+#
+# The branch is on a trait that is constant per type, so exactly one of these is
+# compiled into any given traversal.
+@inline _reusable_children_with_extents(node) =
+    _reusable_children_with_extents(Val(node_extent_is_expensive(node)), node)
```
asinghvi17: dynamic dispatch on node level, tut tut tut...

src/utils/SpatialTreeInterface/dual_depth_first_search.jl
```diff
@@ -1,58 +1,99 @@
...
+) where {F, P, N1, E1, N2, E2}
+    leaf1 = isleaf(node1)
+    leaf2 = isleaf(node2)
+    if leaf1 && leaf2
+        # both nodes are leaves, so we can just iterate over the indices and extents.
+        # `child_indices_extents` is bound once here rather than being re-evaluated
+        # per outer iteration - implementations are encouraged to materialize a
+        # vector there, and that would otherwise be rebuilt for every cell in node1.
+        inner = child_indices_extents(node2)
+        for (i1, cell_extent1) in child_indices_extents(node1)
```
asinghvi17: cie_1 = child_indices_extents(node1)
cie_2 = child_indices_extents(node2)

keep the code clean please

and that comment above is way too long

src/utils/SpatialTreeInterface/interface.jl
```diff
@@ -84,7 +84,37 @@ Generally, defining `Extents.extent(node)` is sufficient here, and you
...
+`node_extent` calls, which is only worth it when the calls are expensive - hence
+the trait rather than doing it unconditionally.
+
+## Implementation notes
+
+For type stability, if your node type is `MyNode`, define
+`node_extent_is_expensive(::Type{MyNode}) = true`, and
+`node_extent_is_expensive(::MyNode)` will forward to that method automatically.
+The value must be inferrable from the type, since it selects between two
+different iteration strategies at compile time.
```
asinghvi17: way too long.  simple DESCRIPTIVE description without talking about the story and whatnot.  just say that if computing node_extent is expensive on this node, tell us and then we cache it.

src/utils/SpatialTreeInterface/README.md
```diff
@@ -29,6 +29,7 @@ These are the only methods that are required to be implemented.
 
 Optionally, one may define:
 - `node_extent(node)` - get the extent of a node.  This falls back to `GI.extent` but can potentially be overridden if you want to return a different but extent-like object.
+- `node_extent_is_expensive(node)::Bool` - whether `node_extent` *computes* the node's extent rather than reading a stored one.  Defaults to `false`, which is right for any tree that stores an extent per node.  Trees that derive extents on access (a DGGS cursor fitting a spherical cap over cell geometry, say) should define it to be `true`, which lets `dual_depth_first_search` derive each child's extent once per node pair instead of once per opposing child.
```
asinghvi17: decrease length of thing

src/utils/SpatialTreeInterface/dual_depth_first_search.jl
asinghvi17: You're right, and it's worse than just the generator being awkward — `Base.Generator{I,F}` with a non-concrete `I` gives a `UnionAll` return type, so the object gets boxed and every `iterate` on it is a dynamic dispatch.

Reproduced with a tree whose nodes hold their children in a tuple, fanouts 2–5 (so `getchild(node)::Tuple{Vararg{Node}}`):

```
_children_with_extents(::Node) :: Base.Generator{I, …} where I<:Tuple{Vararg{Node}}
```

and in the optimized IR of the 6-arg traversal, before/after:

| | dynamic `:call` sites | `@allocated` |
|---|---|---|
| varying fanout, before | 4 (`Base.Generator(...)`) | 45360 B |
| varying fanout, after | 0 | 0 B |
| fixed fanout (`NaturalIndex`), before/after | 0 / 0 | 0 B / 0 B |

Fixed by dropping the pair iterator entirely, as you suggested: loop over `getchild(node)` and call `node_extent(child)` in the body. Worth noting `enumerate` has exactly the same problem (`Enumerate{I} where I`) — I tried it first for the cached path and it put one dynamic dispatch straight back — so the inner loop uses a plain counter instead.

Test suite now has a varying-fanout tree, and the zero-allocation assertion covers it as well as the fixed-fanout case.

src/utils/SpatialTreeInterface/dual_depth_first_search.jl
asinghvi17: Changed, but with a caveat — I couldn't actually reproduce a dynamic dispatch here, so I want to be straight about what the change does.

`Val(node_extent_is_expensive(node))` did const-fold in every case I checked (`NaturalIndex`, the varying-fanout tree, and a tree with the trait set to `true`): 0 dynamic `:call` sites in the optimized IR for all three. Inference constant-propagates the trait because it returns a literal for the type, so the `Val` gets built at compile time.

That said, it only works *because* const-prop succeeds, which isn't something to lean on inside a deeply recursive function — so I've made it not matter:

```julia
@inline function _child_extents(node::N) where {N}
    node_extent_is_expensive(N) || return nothing
    return [node_extent(child) for child in getchild(node)]
end
```

`N` is a static parameter, so the branch is resolved from the type with no const-prop required. Both arms are asserted with `@inferred` in the tests — `nothing` for the default, `Vector{<:Extent}` for an opted-in tree.

src/utils/SpatialTreeInterface/dual_depth_first_search.jl
asinghvi17: Done — `cie_1`/`cie_2` as you wrote it, and the four-line comment is now one line:

```julia
# bound once each - `cie_2` would otherwise be rebuilt per cell of node1
cie_1 = child_indices_extents(node1)
cie_2 = child_indices_extents(node2)
```

src/utils/SpatialTreeInterface/interface.jl
asinghvi17: Cut to three sentences:

> Return true if `node_extent` computes the node's extent instead of reading one the node already stores. Defaults to false.
>
> When true, `dual_depth_first_search` caches a node's child extents rather than re-deriving them once per opposing child, at the cost of a small vector per visited node.
>
> **Implementation notes** — Define this on the type, `node_extent_is_expensive(::Type{MyNode}) = true`, so that it is known at compile time; `node_extent_is_expensive(::MyNode)` forwards there automatically.

Also trimmed the other blocks in these files: the `dual_depth_first_search` docstring paragraph, the comment above the 6-arg method (kept short — the precondition that the extents must equal `node_extent` of their nodes is load-bearing), and, on #463, the `node_size` and `dual_depth_first_search_balanced` docstrings.

src/utils/SpatialTreeInterface/README.md
asinghvi17: Shortened to one line:

> - `node_extent_is_expensive(node)::Bool` - whether `node_extent` computes the extent rather than reading a stored one. Defaults to `false`; set it on trees that derive extents on access, and `dual_depth_first_search` will cache child extents instead of re-deriving them.

Trimmed the allocation note further down the file too.

PR #465, opened 2026-08-13
Make Documenter warnings errors, and fix everything that surfaces

asinghvi17:
Follow-up to #464. That PR fixed the dead links; this one turns off `warnonly = true` so the next regression fails CI instead of quietly degrading the docs, and fixes the backlog that was being swallowed.

`warnonly = true` has been in `docs/make.jl` since #50. Every Documenter check — broken examples, unresolved cross-references, duplicated and missing docstrings — has been a warning ever since, which is how the `@ref` breakage in #464 survived two years and only surfaced as a VitePress dead-link failure.

## What was hiding

18 issues, all of them real. The `@example` failures mean those code blocks currently render as un-expanded source on the docs site, with no output.

**Broken `@example` blocks**

| Where | Problem |
|---|---|
| `spatial_joins.md:17` | `using FlexiJoins,` — trailing comma. The setup block fails to parse, so **all four** later blocks on the page run with nothing imported. One typo, five failures. |
| `geospatial_geometry.md:88` | `poly!(ga, land_features)` passes a whole `DataFrame` — no Makie recipe for that. |
| `geospatial_geometry.md:134` | `DisplayAs.setcontext(y, ...)` — the variable is `ys`. |
| `cap.jl:32` | `using GeometryOps` does not bring `UnitSphericalPoint`/`SphericalCap` into scope; they live in `GeometryOps.UnitSpherical`. |
| `RobustCrossProduct.jl:34` | `⋅` used without importing it. |
| `smooth.jl:32-33` | `GeographicFromUnitSpherical` has never existed — the transform is `GeographicFromUnitSphere`. |
| `segmentize.jl:85` | Not a GEOS failure, despite the `error("LibGEOS: Error in GEOSDensify")` in the visible source. It dies in `benchmarks/benchmark_plots.jl:115`. |

That last one is worth spelling out: current Makie builds `ScatterLines` as `lines!(p, p.attributes, ...)`, so the child `Lines` plot shares the parent's `alpha` node. `setproperty!(child, :alpha, 0.1)` therefore routes to `add_input!` and throws `Cannot attach input with name alpha - already exists!`. Fixed by fading the line through `color` and pinning the markers back with `markercolor`, which keeps the original intent (faint lines, opaque markers).

**Missing and duplicated docstrings**

- `barycentric_coordinates`, `barycentric_coordinates!`, `barycentric_interpolate` are exported public API with **no docstrings at all** — their `@docs` entries in `api.md` have been rendering nothing. Written.
- `circumcenter_on_unit_sphere` was listed in a `@docs` block with no docstring. Written.
- `reproject` and `applyreduce` both carry their docstring on a bare `function f end`, which registers under the `Union{}` signature — exactly what a bare `@docs` entry registers, hence the collision. (Functions like `area` don't collide because their docstrings sit on a method with a concrete signature.) `reproject` is now excluded from the catch-all `@autodocs`; `applyreduce.jl`'s block is marked `canonical=false`, matching what `apply.jl` next door already does.

**Stray Julia interpolation**

Two commented-out lines in `apply.jl`/`applyreduce.jl` become Literate markdown, so `$Target`/`$target` parsed as interpolations. Escaped.

**Unrelated but adjacent**

GLMakie is a docs dependency that nothing uses and that fails to precompile on every single build (`X11: The DISPLAY environment variable is missing`). Dropped.

## Scope

`warnonly = [:missing_docs]` rather than fully empty. The `:missing_docs` check flags 58 docstrings, and every one of them is in an internal submodule — `UnitSpherical`, `SpatialTreeInterface`, `FlexibleRTrees`, `NaturalIndexing`, `LoopStateMachine`, `RobustCrossProduct` — reached because `@autodocs` in `api.md` only sweeps `GeometryOps` and `GeometryOpsCore`, not their submodules. Clearing it means deciding whether things like `_arc_fraction`, `isless_vector` and `normalizableFromExact` belong in the public manual, which is an editorial call that deserves its own PR rather than being rushed to make a flag flip. Everything else is now an error.


benchmarks/benchmark_plots.jl
```diff
@@ -112,7 +112,14 @@ function plot_trials(
             ygridwidth = 0.75,
         ) 
         plots = [scatterlines!(ax, x, y; label = label) for (x, y, label) in zip(xs, ys, labels)]
-        setproperty!.(getindex.(getproperty.(plots, :plots), 1), :alpha, 0.1)
+        # De-emphasize the connecting lines while keeping the markers opaque.  `ScatterLines`
+        # shares a single `alpha` with its child `Lines` plot, so fade the line via `color`
+        # and pin the markers back to the cycled colour with `markercolor`.
```
asinghvi17: delete this comment

benchmarks/benchmark_plots.jl
```diff
@@ -112,7 +112,14 @@ function plot_trials(
...
+        for plot in plots
+            color = Makie.to_color(plot.color[])
+            plot.markercolor = color
+            plot.color = Makie.RGBAf(color.r, color.g, color.b, 0.1)
+        end
```
asinghvi17: Why don't you just specify the colors, and then to get Makie.wong palette or just get the color rotation in the palette, and then set them explicitly when you create the plots.  Don't do this loop after.

docs/make.jl
```diff
@@ -156,7 +156,13 @@ makedocs(;
         ],
         "Source code" => literate_pages,
     ],
-    warnonly = true,
+    # Every Documenter check is an error, so that broken examples, dead cross-references
+    # and duplicated docstrings fail CI instead of silently degrading the built docs.
+    # `:missing_docs` is still only a warning: the docstrings it flags all live in
+    # internal submodules (`UnitSpherical`, `SpatialTreeInterface`, `FlexibleRTrees`,
+    # `NaturalIndexing`, `LoopStateMachine`) which have no API page yet, since the
+    # `@autodocs` blocks in `api.md` only sweep `GeometryOps` and `GeometryOpsCore`.
```
asinghvi17: delete comment

docs/src/api.md
asinghvi17, line 1: add sections for all the submodules here

docs/Project.toml
```diff
@@ -19,11 +19,11 @@ ExactPredicates = "429591f6-91af-11e9-00e2-59fbe8cec110"
 Extents = "411431e0-e8b7-467b-b5e0-f676ba4f2910"
 FlexiJoins = "e37f2e79-19fa-4eb7-8510-b63b51fe0a37"
 GADM = "a8dd9ffe-31dc-4cf5-a379-ea69100a8233"
-GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
 GeoDataFrames = "62cb38b5-d8d2-4862-a48e-6a340996859f"
 GeoDatasets = "ddc7317b-88db-5cb5-a849-8449e5df04f9"
 GeoFormatTypes = "68eda718-8dee-11e9-39e7-89f7f65f511f"
 GeoInterface = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
+GeoInterfaceMakie = "0edc0954-3250-4c18-859d-ec71c1660c08"
```
asinghvi17: this should never be a dep

PR #470, opened 2026-08-14
Add a `Spherical` kernel for `segmentize`

### LINKED ISSUE #461: `segmentize(Spherical(), ...)` throws a MethodError — the kernel was never written
`_segmentize` dispatches on `Union{Planar, Spherical}` and the docstring documents `max_distance` for `Spherical`, but only the `Planar` and `Geodesic` kernels exist.

```julia
import GeometryOps as GO, GeoInterface as GI
ls = GI.LineString([(-50.0, 52.0), (50.0, 52.0)])
GO.segmentize(GO.Spherical(), ls; max_distance = 0.01)
```

```
MethodError: no method matching _fill_linear_kernel!(::GeometryOpsCore.Spherical{Float64}, ::Vector{Tuple{Float64, Float64}}, ::Float64, ::Float64, ::Float64, ::Float64; max_distance::Float64)

Closest candidates are:
  _fill_linear_kernel!(!Matched::GeometryOpsCore.Planar, ::Vector, ::Any, ::Any, ::Any, ::Any; max_distance)
   @ GeometryOps src/transformations/segmentize.jl:220
```

- `src/transformations/segmentize.jl:206` — `_segmentize(method::Union{Planar, Spherical}, ...)`
- `src/transformations/segmentize.jl:220` — `_fill_linear_kernel!(::Planar, ...)`, the only one in `src`
- `ext/GeometryOpsProjExt/segmentize.jl:58` — `_fill_linear_kernel!(::Geodesic, ...)`

`git log -S _fill_linear_kernel!` across every ref shows a `Spherical` method has never existed on any branch, so this is a gap rather than a regression. Version: v0.1.43 (`main`).

### Fix

`slerp` is already there. Roughly:

```julia
function _fill_linear_kernel!(m::Spherical, new_coords::Vector, x1, y1, x2, y2; max_distance)
    a = UnitSphereFromGeographic()((x1, y1))
    b = UnitSphereFromGeographic()((x2, y2))
    distance = spherical_distance(a, b) * m.radius
    if distance > max_distance
        n_segments = ceil(Int, distance / max_distance)
        for i in 1:(n_segments - 1)
            push!(new_coords, GeographicFromUnitSphere()(slerp(a, b, i / n_segments)))
        end
    end
    push!(new_coords, (x2, y2))  # exact, like the planar kernel
    return nothing
end
```

Per the docstring, `max_distance` is in units of the radius — arc length `Ω * m.radius` — so metres under the default radius and radians under `Spherical(; radius = 1)`.

Worth a test that a segmentized parallel bows poleward of the parallel itself, since that is the entire difference from `Planar`: the midpoint of `(-50, 52)`–`(50, 52)` is `slerp` → 63.33°N, i.e. 11.33° off the parallel.

asinghvi17:
Closes #461.

`_segmentize` dispatches on `Union{Planar, Spherical}` and the docstring documents `max_distance` for `Spherical`, but only the `Planar` (in `src`) and `Geodesic` (in `GeometryOpsProjExt`) kernels were ever written — so `segmentize(Spherical(), geom; max_distance)` threw a `MethodError`.

```julia
julia> import GeometryOps as GO, GeoInterface as GI

julia> GO.segmentize(GO.Spherical(), GI.LineString([(-50.0, 52.0), (50.0, 52.0)]); max_distance = 3_500_000)
```

Before: `MethodError: no method matching _fill_linear_kernel!(::Spherical{Float64}, ...)`.
After:

```
GeoInterface.Wrappers.LineString([(-50.0, 52.0), (-7.087e-15, 63.334), (50.0, 52.0)])
```

## What's here

- **`_fill_linear_kernel!(::Spherical, ...)`** — walks the great circle joining each pair of points via `slerp`, which was already in `UnitSpherical`. It mirrors the `Planar` kernel: `ceil(Int, distance / max_distance)` equal segments, and the segment ends on the *original* coordinate rather than an interpolated one.

  `max_distance` is an arc length in units of `Spherical.radius` — the subtended angle `Ω` times the radius — so it's metres under the default (mean Earth) radius and radians under `Spherical(; radius = 1)`. That's what the existing docstring already promised.

- **A fix for the three-argument form.** `segmentize(method::Manifold, geom, max_distance)` dropped `method` and always called through with `Planar()`, so `segmentize(Spherical(), geom, 1000)` silently returned a planar segmentization. Noticed while writing the tests for the above; it's a one-word fix with a test.

- **Docs.** `Spherical` is added to the manifold list in the literate header and to the `segmentize` docstring's `method` argument, with a worked example of the poleward bow and a note that `max_distance` is in radius units.

## Tests

The interesting property is the one the issue calls out: the great circle between two points on the same parallel bows *poleward* of that parallel, so a spherical segmentization is not a refinement of the planar one. The midpoint of `(-50, 52)`–`(50, 52)` lands at `atand(tand(52) / cosd(50))` ≈ 63.33°N, 11.33° off the parallel.

Beyond that: endpoints are preserved exactly, `npoint` matches `ceil(Ω / max_distance) + 1`, results are invariant under scaling `radius` and `max_distance` together, no output segment exceeds `max_distance`, short and zero-length arcs are left alone, and rings stay closed. The existing `@testset_implementations` blocks now cover `Spherical` alongside `Planar` and `Geodesic`, and the test file imports LibGEOS/ArchGDAL/GeometryBasics so those macros actually run against all four implementations rather than GeoInterface alone.

Full local test suite passes (`Segmentize` 266/266).

### DIFF
--- a/src/transformations/segmentize.jl
+++ b/src/transformations/segmentize.jl
@@ -186,7 +186,8 @@ This is useful for plotting geometries with a limited number of vertices, or for
 ## Arguments
 - `method::Manifold = Planar()`: The method to use for segmentizing the geometry.  At the moment, [`Planar`](@ref) (assumes a flat plane), [`Spherical`](@ref) (assumes geometry on a sphere and interpolates along great circles) and [`Geodesic`](@ref) (assumes geometry on the ellipsoidal Earth and uses Vincenty's formulae) are available.
 - `geom`: The geometry to segmentize.  Must be a `LineString`, `LinearRing`, `Polygon`, `MultiPolygon`, or `GeometryCollection`, or some vector or table of those.
-- `max_distance::Real`: The maximum distance between vertices in the geometry.  **Beware: for `Planar`, this is in the units of the geometry, but for `Geodesic` and `Spherical` it's in units of the radius of the sphere.**  So under the default `Spherical()` radius `max_distance` is in metres, and under `Spherical(; radius = 1)` it is in radians.
+- `max_distance::Real`: The maximum distance between vertices in the geometry.  
+  For `Planar` manifolds, this is in the units of the geometry.  For `Spherical` and `Geodesic`, this is in the units of the manifold (the units of the radius of the sphere/ellipsoid).  By default this is meters.
 
 `Spherical` and `Geodesic` both assume that the input geometry is in lon/lat coordinates, in degrees.
 
--- a/src/transformations/segmentize.jl
+++ b/src/transformations/segmentize.jl
@@ -47,13 +47,28 @@ axislegend(a; position = :lt)
 f
 ```
 
-There are two methods available for segmentizing geometries at the moment, 
+There are three methods available for segmentizing geometries at the moment,
 and you can invoke them by passing the relevant [`Manifold`](@ref):
 
 ```@docs; canonical=false
 Planar
+Spherical
 Geodesic
 ```
+
+[`Spherical`](@ref) interpolates along the great circle joining each pair of points, so it
+sits between [`Planar`](@ref) and [`Geodesic`](@ref): it accounts for the curvature of the
+Earth, but treats it as a sphere rather than an ellipsoid.  A spherical segmentization is
+*not* a refinement of the planar one — the great circle joining two points on the same
+parallel bows poleward of that parallel:
+
+```@example segmentize
+parallel = GI.LineString([(-50.0, 52.0), (50.0, 52.0)])
+collect(GI.getpoint(GO.segmentize(GO.Spherical(), parallel; max_distance = 3_500_000)))
+```
+
+The added midpoint sits at 63.3°N, over 11° north of the 52°N parallel that `Planar` would
+have interpolated along.
 
 ## Benchmark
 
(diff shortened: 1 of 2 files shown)


src/transformations/segmentize.jl
```diff
@@ -169,9 +184,11 @@ Segmentize a geometry by adding extra vertices to the geometry so that no segmen
...
+- `method::Manifold = Planar()`: The method to use for segmentizing the geometry.  At the moment, [`Planar`](@ref) (assumes a flat plane), [`Spherical`](@ref) (assumes geometry on a sphere and interpolates along great circles) and [`Geodesic`](@ref) (assumes geometry on the ellipsoidal Earth and uses Vincenty's formulae) are available.
 - `geom`: The geometry to segmentize.  Must be a `LineString`, `LinearRing`, `Polygon`, `MultiPolygon`, or `GeometryCollection`, or some vector or table of those.
-- `max_distance::Real`: The maximum distance between vertices in the geometry.  **Beware: for `Planar`, this is in the units of the geometry, but for `Geodesic` and `Spherical` it's in units of the radius of the sphere.**
+- `max_distance::Real`: The maximum distance between vertices in the geometry.  **Beware: for `Planar`, this is in the units of the geometry, but for `Geodesic` and `Spherical` it's in units of the radius of the sphere.**  So under the default `Spherical()` radius `max_distance` is in metres, and under `Spherical(; radius = 1)` it is in radians.
```
asinghvi17:
```suggestion
- `max_distance::Real`: The maximum distance between vertices in the geometry.  
  For `Planar` manifolds, this is in the units of the geometry.  For `Spherical` and `Geodesic`, this is in the units of the manifold (the units of the radius of the sphere/ellipsoid).  By default this is meters.
```

Issue #469, opened 2026-08-14
RelateNG prepared mode rebuilds the B-side geometry every call: ~35 kB/call in a one-target x many-geometries sweep


asinghvi17:
`prepare(RelateNG(), a)` caches the **A** side and `relate_predicate(prep, pred, b)` reuses it — that part works, and it is what we use downstream. But `evaluate!` builds a fresh `RelateGeometry` for **`b`** on every call:

https://github.com/JuliaGeo/GeometryOps.jl/blob/4e0c0d5199a706889bb96714d667db39e59a88c4/src/methods/geom_relations/relateng/relate_ng.jl#L272-L280

and there is no hook to reuse that work or its buffers across calls. In the one-target-versus-many-geometries sweep that prepared mode exists to serve, the B side is then the **entire** remaining cost.

**Measured** on `main` @ 4e0c0d5 (Apple M-series, Julia 1.11), one prepared `Spherical()` polygon target, `pred_contains` (the converse of a `Within` query). Full script in a comment below.

Per-call cost is flat in the *target's* size — prepared mode is doing its job:

| target vertices | 4 | 32 | 72 | 312 | 616 | 736 |
| --- | --- | --- | --- | --- | --- | --- |
| KiB/call | 7.5 | 7.5 | 7.5 | 7.5 | 7.5 | 7.5 |
| us/call | 8.35 | 8.38 | 8.54 | 8.19 | 8.30 | 8.00 |

and linear in the *query geometry's* size, at ~1 KiB and ~0.8 us per vertex:

| query vertices | 3 | 6 | 16 | 32 | 64 | 128 |
| --- | --- | --- | --- | --- | --- | --- |
| KiB/call | 4.5 | 7.5 | 17.7 | 33.8 | 66.1 | 132.0 |
| us/call | 6.23 | 7.86 | 15.29 | 27.79 | 55.50 | 102.75 |

For scale, `prepare` on those same targets costs 7.3 KiB / 13 us (4 verts) up to 899.5 KiB / 645 us (736 verts) — so against a large target a single 128-vertex query call is already ~15% of a full re-preparation, and against a small one it is several times the preparation.

A **point** query is the one cheap case, at 0.1 KiB / 0.01 us. That is a query-shape short-circuit, not a predicate one: `pred_contains`, `pred_within` and `pred_intersects` on the same 6-vertex ring all cost an identical 7.5 KiB.

[...]

PR #472, opened 2026-08-14
Reduce the per-call B-side cost of prepared `RelateNG` queries

asinghvi17:
Part of #469 — the per-call B-side rebuild in prepared mode.

## What this does

Profiling one prepared `relate_predicate(prep, pred_contains(), cell)` call on the issue's scenario (Spherical, 72-vertex prepared target, 32-vertex query cell, fresh predicate per call — note the issue's benchmark loop reuses one already-determined predicate, so its tables understate the true per-call cost, which is **85.5 KB / 90 µs** on `main`) shows the cost is not mainly the B data structures themselves:

- **~31 KB / 25 µs (91% of the B `RelateGeometry` build)**: the B polygon's region extent — `_spherical_region_extent` runs six axis-point enclosure queries through `spherical_ring_contains`, each O(n²) with allocating float predicates, on every call.
- **~34 KB**: `Rational{BigInt}`-lifted composite predicates in point location under `exact = True()` (not touched here; wants an S4-style float-filter ladder as follow-up).
- **2.6 KB / 0.6 µs**: the per-call RTree over B's segments, built unconditionally whenever A has a prepared tree.

Three changes:

1. **Vertex-cap prefilter for the region-extent axis queries** (`extent.jl`). If all ring vertices fit in a sub-hemisphere cap, the ring's arcs stay in the (geodesically convex) cap, so the cap's complement is boundary-free and containment is uniform across it: one probe query — taken orthogonal to the vertex mass, away from the near-antipodal configurations that push `robust_cross_product` onto its exact fallback — answers for every axis point outside the cap. Small rings pay one O(n²) query instead of six.
2. **Stored-extent reuse on `Spherical`** (`kernel_spherical.jl`, `relate_geometry.jl`). `rk_interaction_bounds(::Spherical)` now returns a stored `(X, Y, Z)` extent as-is, exactly as the planar kernel already trusts any stored extent, and the polygon wrapper built by `_relate_cache_extents` stores the region box (rather than the union of ring boxes) so the shortcut is sound. A B stamped with kernel-space extents at every level — e.g. the wrapper tree a previous `RelateGeometry` built — skips the extent recomputation entirely.
3. **No per-call B tree for small B** (`relate_ng.jl`). Below the existing accelerator threshold, each B segment is a single-tree query into the prebuilt A tree instead of building a fresh RTree over B per call.

## Measured (Apple M-series, Julia 1.12, fresh `pred_contains` per call)

| configuration | `main` @ 4e0c0d5 | this PR |
|---|---|---|
| raw B, default | 85.5 KB / 90 µs | 73.5 KB / 71 µs |
| raw B, `exact = False()` | 51.6 KB / 40 µs | 39.5 KB / 30 µs |
| pre-stamped B, default | no benefit | 50.6 KB / 26 µs |
| pre-stamped B, `exact = False()` | — | 16.7 KB / 8.8 µs |
| B `RelateGeometry` build alone | 34.5 KB / 27.5 µs | 22.5 KB / 15.4 µs raw · 176 B / 0.02 µs pre-stamped |

## Correctness

- Full RelateNG suite passes: 2202 engine, 6537 JTS XML conformance, 4022 S2 crossing conformance, 500 LibGEOS differential fuzz, allocation/type-stability.
- All 908 spherical extent tests pass (they cover the cap prefilter's tricky cases: dumbbell rings with zero net winding, super-hemisphere caps, CW complements under `oriented = true`, random cells vs axis points).
- New tests: prepared small-B and at-threshold paths against the unprepared engine (including straddling cells, so edge topology flows through the single-tree query path), stored-3D-extent reuse (wrapper identity + extent identity + result equality), and a wrong-space lon/lat stored extent being ignored.
- The issue's 441-cell MWE classifies identically before and after.

## Not addressed here (follow-ups for #469)

- The `BigInt` lifting in point location under `exact = True()` (~34 KB/call) — float-filter-first ladder, as was done for `rk_classify_intersection`.
- The residual scratch (segment extraction, locator shells, node dicts, ~17 KB) — the issue's proposed `workspace` argument.
- A public API for stamping a B geometry with kernel-space extents (today only internal wrappers get the fast path).

### DIFF
--- a/src/methods/extent.jl
+++ b/src/methods/extent.jl
@@ -91,11 +91,51 @@ function _spherical_region_extent(pts::Vector{<:UnitSpherical.UnitSphericalPoint
     end
     n < 3 && return ext
 
+    #=
+    Vertex-cap prefilter for the six axis queries below. If every vertex lies
+    within angle θ < 90° of the normalized vertex mass `c`, the ring's minor
+    arcs also stay inside the (geodesically convex) cap around `c` — for arc
+    points `w = α u + β v` with `α, β ≥ 0` and `α + β ≥ 1` (concavity of sin),
+    `w ⋅ c ≥ min(u ⋅ c, v ⋅ c)` — so the cap's complement is connected and
+    boundary-free, and containment is uniform across it: one query at a
+    probe point outside the cap answers for every axis point outside it. A
+    small ring (the common case) pays one O(n²) containment query instead
+    of six. The `cos θ > 0` requirement also excludes (near-)antipodal
+    edges, whose arc choice the cap bound would not survive. The probe is
+    taken orthogonal to `c` — outside any θ < 90° cap, and far from the
+    near-antipodal configurations (probe vs edge midpoints ≈ `c`) that
+    would push `robust_cross_product` onto its exact fallback.
+    =#
+    mass = pts[1]
+    for i in 2:n
+        mass += pts[i]
+    end
+    nc = norm(mass)
+    cap_ok = nc > 1e-6 * n
+    c = cap_ok ? UnitSpherical.UnitSphericalPoint(mass / nc) : first(pts)
+    costheta = 1.0
+    if cap_ok
+        for i in 1:n
+            costheta = min(costheta, pts[i] ⋅ c)
+        end
+        cap_ok = costheta > 1e-6
+    end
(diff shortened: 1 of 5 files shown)


asinghvi17:
Looks good to me

PR #473, opened 2026-08-19
Add an `intersection_area` prototype

asinghvi17:
`intersection_area(alg, a, b; cache)` computes `area(manifold(alg), intersection(alg, a, b))` without building the intersection geometry — each engine sums the rings it would otherwise have wrapped in a `Polygon`.

Implemented for all three clipping engines. The supported-algorithm list is closed — there is deliberately no algorithm-free fallback, since a prototype whose point is "without constructing the geometry" shouldn't quietly fall back to constructing it.

| engine | manifolds | how |
|:---|:---|:---|
| `ConvexConvexSutherlandHodgman` | `Planar()`, `Spherical()` | measures the clip buffers in place; with a `cache`, allocation-free |
| `OverlayNG` | `Planar()`, `Spherical()` | ring-area sum in place of polygon assembly |
| `FosterHormannClipping` | `Planar()` | accumulates during the trace, so no ring is materialized |

**Where the code lives.** The public file is only the export, docstring and dispatch; the work sits with each engine:

- `_ring_area` / `_area_scale` in `methods/area.jl`, next to the `_signed_area` and `_naive_triangulated_spherical_ring_area` formulas they have to stay term-for-term equal to — that equality is what makes the two paths agree bitwise, so they are commented as a pair;
- `_sh_clip_planar!` / `_sh_clip_spherical!` in `sutherland_hodgman.jl`;
- `_overlay_intersection_area` beside the driver in `overlay_ng.jl`, `_build_polygon_area` beside `_ring_to_polygon` in `polygon_builder.jl`;
- `_RingCollector` / `_RingMeasurer` in `clipping_processor.jl`.

Supporting refactors, all behaviour-preserving: `_build_polygons` splits into `_build_polygon_ctx` + the polygon wrap; the middle of `_overlay_ng` becomes `_overlay_marked_graph`; a shared `_overlay_input` builds the engine's input model so every entry point derives dimensions and empty flags the same way, and returns the dimensions alongside it (see below); `_trace_polynodes` takes a ring sink instead of always building a point vector.

**Foster-Hormann scope**: the streamed path covers hole-free polygon pairs. Holes and multipolygons are assembled from the traced rings *afterwards* — `_add_holes_to_polys!` cuts holes into result polygons and can split, merge and drop them — and there is no ring-level shortcut through that, so those inputs delegate to the polygon path. Same answer, no speedup.

**Measured**, mean over a loop that consumes the result (BenchmarkTools eliminates the allocation-free call), on the 2x2 square pair from the docstring:

| | time | bytes | allocs |
|---|---|---|---|
| `area ∘ intersection`, SH + cache | 141 ns | 288 | 7 |
| `intersection_area`, SH + cache | 86 ns | 0 | 0 |
| `area ∘ intersection`, FH | 1270 ns | 2848 | 27 |
| `intersection_area`, FH | 978 ns | 2096 | 14 |
| `area ∘ intersection`, OverlayNG | 4358 ns | 15856 | 254 |
| `intersection_area`, OverlayNG | 4103 ns | 15392 | 240 |

OverlayNG gains least — noding dominates, and polygon assembly was never the cost. FH's remaining allocations are the node lists from `_build_ab_list`.

**Behaviour change worth flagging**: the spherical Sutherland-Hodgman containment fallback previously returned a `Polygon` around the clip ring at any length, so a degenerate `< 3`-point clip polygon inside the subject produced an invalid ring. The `length(pts) < 3` guard now covers that path too.

**Tests**: `test/methods/clipping/intersection_area.jl` (78 tests) covers all three engines against `area(intersection(...))`, holes on both manifolds, disconnected results, multipolygons, sub-areal inputs, the lon/lat output chart, `Float32` return types, radius scaling, trait rejection, and the allocation-free cached path. The existing Sutherland-Hodgman suite (85 tests) still passes. Every entry point is `@inferred`-clean.

Fuzzing: 500 random polygon pairs per engine/manifold, plus a sliver-heavy sweep of 2000 pairs whose overlaps are corner slivers with widths from 1e-4 down to 1e-14. `intersection_area` is **bit-identical** to `area(intersection(...))` in every case, and the four-op OverlayNG area identities hold to 4e-16 (planar) / 3e-15 (spherical).

The sliver sweep exists because the first version of this PR was wrong on exactly those inputs, and the random-convex fuzz never generated them: `_ring_area(::Spherical, ...)` trimmed a trailing point that was `≈` the first, but the Sutherland-Hodgman buffers are open rings, so on a sliver whose two end vertices fell inside `isapprox`'s default rtol (~1.5e-8) it ate a real vertex — halving the area, or zeroing it once the remaining fan degenerated. `_ring_area` now takes `closed::Bool = true` to gate the trim, and the Sutherland-Hodgman paths pass `false`; OverlayNG's rings are closed (`ring[1] === ring[end]`) and keep the default. No tolerance can substitute for the flag: the two ends of a sliver are legitimately that close together.

**Three pre-existing bugs found along the way** (neither introduced here, neither fixed here — `intersection.jl` and the FH constructors are untouched):

1. `GO.intersection(a, b)` with no `target` throws `MethodError: no method matching TraitTarget(::Nothing)`. Every call form that omits `target` hits it, including the algorithm-free one. `intersection_area`'s FH fallback names `target = GI.PolygonTrait()` to route around it.
2. `GO.FosterHormannClipping(GO.Spherical())` is an ambiguous constructor call (`clipping_processor.jl:52` vs `:62`), so no spherical FH algorithm can be constructed. The spherical branch of the FH ring measurer is written but consequently unreachable and untested, and marked as such in a comment.
3. `area(::Spherical, geom)` trims its ring's closing point with the same `≈`, and `area` documents that an unclosed curve "is still assumed to be closed", so an unclosed sliver ring gets the same wrong answer through the public function. Not fixed here: it changes a core public function and belongs in its own PR.

**Full suite run.** Every suite this touches was run against `main` on the same machine, same manifest:

| | this branch | `main` |
|:---|:---|:---|
| totals | 13893 pass / 222 fail / 20 error / 8 broken | 13815 pass / 222 fail / 21 error / 8 broken |
| `methods/area.jl` | 135 pass | 135 pass |
| `polygon_clipping.jl` | 6347 pass | 6347 pass |
| `sutherland_hodgman.jl` | 85 pass | 85 pass |
| `intersection_area.jl` | 78 pass | — |
| OverlayNG engine | 765 pass / 4 error | 765 pass / 4 error |
| OverlayNG `xml_suite.jl` | 522 pass / 221 fail / 1 error / 2 broken | identical |
| OverlayNG `fuzz.jl` | 737 pass / 1 fail / 6 error | identical |
| noding, graph, faces, points, API, labeller, identities, s2_differential | unchanged | unchanged |

The delta is exactly +78 (the new tests) and −1 (`main` errors on the file not existing). The 222 failures and 20 errors are all pre-existing on `main` and untouched here — the errors are a `GI.npoint`-over-an-empty-collection `ArgumentError`, the XML-suite failures are the known JTS robust-overlay defects.

**That run caught a real regression in this PR**, since fixed. Reading `input.dim_a` back off `_overlay_input`'s result widened the *untargeted* `intersection` return type from its 7-member `Union` to `Any`: `_OverlayInput` was mutable and escapes into the labeller, so a field load off it is opaque to inference, where `_overlay_dimension(a)` folds to a constant that prunes the union. `@inline` does not rescue it — the escape defeats SROA either way. Fixed properly in 2825c4766 by typing the struct: `a`/`b` became type parameters, the lazy locators became concretely-typed `Ref`s (`IndexedPointInAreaLocator{M,E}` is parameterized only on manifold and exactness, so the `Ref` type is known without building the locator), and the struct became immutable — which both restores the const-fold on a plain field load and makes `locate` a static dispatch in `_input_locate_in_area` (~163ns → ~68ns per call). Those four `Any` fields predate this PR. Only the three inference assertions in "target makes the return type concrete and input-independent" ever failed; no result changed.

https://claude.ai/code/session_01DZi2npxCab69YAUv6qYk9a

### DIFF
--- a/src/methods/clipping/clipping_processor.jl
+++ b/src/methods/clipping/clipping_processor.jl
@@ -919,6 +919,61 @@ function _index_crossing_intrs!(alg::FosterHormannClipping{M, A}, a_list, b_list
     return
 end
 
+# Get type of polygons that will be made
+# TODO: Increase type options
+_get_poly_type(::Type{T}) where T =
+    GI.Polygon{false, false, Vector{GI.LinearRing{false, false, Vector{Tuple{T, T}}, Nothing, Nothing}}, Nothing, Nothing}
+
+#= What the tracer does with the vertices it walks is the only thing that separates
+building the result rings from measuring them, so that is the one thing it takes as a
+parameter. `_RingCollector` is the original behaviour - a point vector per ring, wrapped
+as a polygon. `_RingMeasurer` keeps a running ring area instead, which is what lets
+`intersection_area` trace without materializing a ring at all.
+
+Each sink owns a per-ring `state`: `_ring_start` opens a ring at its first vertex,
+`_ring_step` takes the next vertex and returns the new state, and `_ring_close!` folds the
+finished ring into the sink. =#
+struct _RingCollector{P}
+    polys::Vector{P}
+end
+_RingCollector(::Type{T}) where {T} = _RingCollector(Vector{_get_poly_type(T)}(undef, 0))
+
+_ring_start(::_RingCollector, pt) = [pt]
+_ring_step(::_RingCollector, pts, pt) = (push!(pts, pt); pts)
+_ring_close!(sink::_RingCollector, pts) = (push!(sink.polys, GI.Polygon([pts])); nothing)
+
+mutable struct _RingMeasurer{M <: Manifold, T}
+    manifold::M
+    area::T
+    nrings::Int   # `isempty(polys)` for the collector: whether the trace found anything
+end
+_RingMeasurer(m::M, ::Type{T}) where {M, T} = _RingMeasurer{M, T}(m, zero(T), 0)
+
+#-- state is (first vertex, previous vertex, running sum): both formulas below are
+#-- two-point recurrences, so the ring never has to exist all at once
+_ring_start(sink::_RingMeasurer{M, T}, pt) where {M, T} = (pt, pt, zero(T))
+_ring_step(sink::_RingMeasurer, (first_pt, prev, acc), pt) =
+    (first_pt, pt, acc + _ring_term(sink.manifold, first_pt, prev, pt))
+function _ring_close!(sink::_RingMeasurer, (first_pt, prev, acc))
+    sink.area += abs(_ring_total(sink.manifold, acc + _ring_term(sink.manifold, first_pt, prev, first_pt)))
+    sink.nrings += 1
+    return nothing
+end
+
+#-- the per-vertex terms of `_ring_area`'s two formulas (methods/area.jl), taken one
+#-- vertex at a time. Summed in the same order, they give the same answer. The spherical
+#-- one is untested: `FosterHormannClipping(Spherical())` is an ambiguous constructor call
+#-- today, so no spherical FH algorithm can be built to reach it.
+_ring_term(::Planar, first_pt, prev, pt) = _area_component(prev, pt)
+_ring_term(::Spherical, first_pt, prev, pt) = _spherical_triangle_area(Eriksson(),
+    UnitSphericalPoint(GI.PointTrait(), first_pt), UnitSphericalPoint(GI.PointTrait(), prev),
+    UnitSphericalPoint(GI.PointTrait(), pt))
+_ring_total(::Planar, acc) = acc / 2
+_ring_total(::Spherical, acc) = acc
+
+_trace_polynodes(alg::FosterHormannClipping, ::Type{T}, a_list, b_list, a_idx_list, f_step, poly_a, poly_b) where {T} =
+    _trace_polynodes!(_RingCollector(T), alg, T, a_list, b_list, a_idx_list, f_step, poly_a, poly_b).polys
+
 #=
     _trace_polynodes(::Type{T}, a_list, b_list, a_idx_list, f_step)::Vector{GI.Polygon}
 
@@ -937,11 +992,10 @@ A list of GeoInterface polygons is returned from this function.
 Note: `poly_a` and `poly_b` are temporary inputs used for debugging and can be removed
 eventually.
 =#
-function _trace_polynodes(alg::FosterHormannClipping{M, A}, ::Type{T}, a_list, b_list, a_idx_list, f_step, poly_a, poly_b) where {T, M, A}
+function _trace_polynodes!(sink, alg::FosterHormannClipping{M, A}, ::Type{T}, a_list, b_list, a_idx_list, f_step, poly_a, poly_b) where {T, M, A}
     n_a_pts, n_b_pts = length(a_list), length(b_list)
     total_pts = n_a_pts + n_b_pts
     n_cross_pts = length(a_idx_list)
-    return_polys = Vector{_get_poly_type(T)}(undef, 0)
     # Keep track of number of processed intersection points
     visited_pts = 0
     processed_pts = 0
@@ -959,7 +1013,7 @@ function _trace_polynodes(alg::FosterHormannClipping{M, A}, ::Type{T}, a_list, b
 
         # Set first point in polygon
         curr = curr_list[idx]
-        pt_list = [curr.point]
+        ring = _ring_start(sink, curr.point)
 
         curr_not_start = true
         while curr_not_start
@@ -975,9 +1029,9 @@ function _trace_polynodes(alg::FosterHormannClipping{M, A}, ::Type{T}, a_list, b
                 idx = (idx > curr_npoints) ? mod(idx, curr_npoints) : idx
                 idx = (idx == 0) ? curr_npoints : idx
 
-                # Get current node and add to pt_list
+                # Get current node and add to the ring
                 curr = curr_list[idx]
-                push!(pt_list, curr.point)
+                ring = _ring_step(sink, ring, curr.point)
                 if (curr.crossing || curr.endpoint != not_endpoint)
                     # Keep track of processed intersection points
                     same_status = curr.ent_exit == prev_status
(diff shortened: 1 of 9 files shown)


src/methods/clipping/intersection_area.jl
```diff
@@ -0,0 +1,134 @@
...
+
+# ## Sutherland-Hodgman
+#
+# The clip buffers are the result; there is nothing left to do but measure them.
+
+function intersection_area(
+    alg::ConvexConvexSutherlandHodgman{Planar}, geom_a, geom_b, ::Type{T}=Float64;
+    cache::Union{Nothing, SutherlandHodgmanCache}=nothing
+) where {T<:AbstractFloat}
+    cache = isnothing(cache) ? SutherlandHodgmanCache(Planar(), T) : _sh_check_cache(cache, Tuple{T,T})
```
asinghvi17: You may as well just create the cache in the kwarg and call the check function internally, then

asinghvi17: Done in 6075c27 — both SH methods now read:

```julia
    cache::SutherlandHodgmanCache = SutherlandHodgmanCache(Planar(), T)
) where {T<:AbstractFloat}
    _sh_check_polygon_traits(GI.trait(geom_a), GI.trait(geom_b))
    _sh_check_cache(cache, Tuple{T,T})
```

The default is only evaluated when no cache is passed, so the cached path still allocates nothing. One knock-on: `intersection_area(alg, a, b; cache = nothing)` is now a `TypeError` rather than meaning "no cache" — `intersection` still accepts it, since I left that signature alone. Say the word if you'd rather the two matched.

src/methods/clipping/clipping_processor.jl
```diff
@@ -919,6 +919,61 @@ function _index_crossing_intrs!(alg::FosterHormannClipping{M, A}, a_list, b_list
...
+_ring_step(::_RingCollector, pts, pt) = (push!(pts, pt); pts)
+_ring_close!(sink::_RingCollector, pts) = (push!(sink.polys, GI.Polygon([pts])); nothing)
+
+mutable struct _RingMeasurer{M <: Manifold, T}
+    manifold::M
+    area::T
+    nrings::Int   # `isempty(polys)` for the collector: whether the trace found anything
+end
+_RingMeasurer(m::M, ::Type{T}) where {M, T} = _RingMeasurer{M, T}(m, zero(T), 0)
+
```
asinghvi17: Worth defining an abstract supertype + formal interface, even if these are not public api.

asinghvi17: Done in a7b3375. `_RingSink` is now an abstract supertype with the interface written out — a table of the three methods, what each returns, and what `state` is (per-ring, tracer-created, never inspected by it; the sink is the mutable half that accumulates across rings). Fallbacks on the abstract type name the missing method:

```
ArgumentError: BadSink is a `_RingSink` but does not implement `_ring_start(sink, pt)`. A ring sink must implement ...
```

rather than failing as a `MethodError` several frames into the traversal. `_trace_polynodes!` dispatches on `::_RingSink`.

Writing the contract down also removed a line: it made explicit that the tracer closes each ring on the vertex it opened with, so the measurer's explicit closing term was always zero. Dropped it — 500 random pairs still give bit-identical areas to the polygon path, and the tracer's own output is still byte-identical to the pre-refactor baseline across 400 pairs × intersection/union/difference.

Kept the `_` prefix and the `#= =#` doc block rather than a docstring, since `docs/src/api.md` ends in an `@autodocs` over the whole module and a docstring would publish it.

asinghvi17:
You're right, and my bit-identity claim was wrong — fixed in 12428d3, exactly as you specified: `closed::Bool = true` on `_ring_area` gating the trim, `false` from both Sutherland-Hodgman paths. OverlayNG keeps the default; I checked its rings are genuinely closed (`ring[1] === ring[end]`), not just approximately.

Reproduced before fixing, and it is worse than a last-ULP disagreement. A corner-overlap sliver of side `w` against a 2°x2° cell:

| w | `intersection_area` | `area(intersection(...))` |
|---|---|---|
| 1e-6 | 0.012356813812376343 | 0.012356813812376343 |
| 1e-8 | 6.174641379391817e-7 | 1.235680613611967e-6 |
| 1e-9 | 6.174668701289166e-9 | 1.2356863829879645e-8 |
| 1e-10 | 6.17470538070696e-11 | 1.2356955829391982e-10 |

Half, not a rounding difference — the trim turns the 4-vertex sliver into a triangle, dropping one of the two fan triangles. Your 1e-12 sr scale and the collapses to 0.0 both follow.

Why my fuzz missed it: it drew random convex polygons at O(1) size, which essentially never produce an overlap whose *first and last buffer vertices* land within 1.5e-8 of each other. The tolerance is `isapprox`'s default rtol, which is enormous in unit-sphere terms — and no tolerance can work here, because the two ends of a sliver are legitimately that close. Only the caller knows whether the vector it passed repeats its first point, which is what the flag now encodes.

Added regression tests: the corner sliver at w in {1e-6, 1e-8, 1e-9, 1e-10, 1e-12} asserting `==` against `area(intersection(...))` and `> 0`, plus a direct test that `closed = false` keeps every vertex of an open near-closed ring while the default still trims a closed one. A new sliver-heavy sweep — 2000 pairs, widths from 1e-4 to 1e-14, spanning the tolerance — is now bit-identical on both manifolds with none collapsing to zero; on the old code the same sweep is where the halving lives. The existing sweeps are unchanged bitwise (the OverlayNG four-op identities still come back at exactly 4.4140389852711566e-16 planar / 2.9941100465588843e-15 spherical).

---

Related, and **not** fixed here because it is a change to `area` itself: `_naive_triangulated_spherical_ring_area` has the same `≈` trim, and `area` documents that an unclosed curve "is still assumed to be closed", so the bug is reachable through the public function:

```julia
open_ring  = [t((0.0, 0.0)), t((1.0, 0.0)), t((1.0, 1.0)), t((1e-9, 1e-9))]
GO.area(GO.Spherical(), GI.Polygon([open_ring]))                      # 6.18248680192442e9
GO.area(GO.Spherical(), GI.Polygon([push!(copy(open_ring), open_ring[1])]))  # 6.182486801923502e9
```

Same shape of failure, and on a polygon that *is* a sliver it would be the same ~2x. Planar is unaffected (it wraps rather than trims). I did not touch it: it changes a core public function, the choice between tightening `≈` to `==` and threading a `closed` flag through has consequences for rings whose closing point is only approximately equal after reprojection, and I cannot run `test/methods/area.jl` here (needs LibGEOS). Happy to do it as a follow-up if you want it, either way.

PR #476, opened 2026-08-21
Strip the discarded work out of two spherical predicates (`spherical_orient`, `SphericalCap` overlap)

asinghvi17:
Two spherical predicates were paying for arithmetic they then threw away. This
strips that work out without changing what either predicate answers.

**Stacked on #473** (`as/intersection_area`) — this PR's base is that branch, so
the diff below is only the two commits on top of it. Please merge #473 first.

### 1. `spherical_orient`: drop the discarded `normalize` (and the asserts that came with it)

`spherical_orient` needs only the *sign* of `n ⋅ c`. It obtained `n` from
`robust_cross_product`, which **normalizes** its result — a `sqrt` and three
divides — and, on the way in, runs two `@boundscheck @assert isUnitLength(...)`
checks on its arguments.

Normalizing `n` scales it by a positive number, so it cannot change the sign of
`n ⋅ c`: the `sqrt`, the divides, and the input asserts are all work the
predicate discards. The common case is now the bare stable cross product
`cross(a - b, a + b)`, written out componentwise and left unnormalized. The
degeneracy band is preserved in form and in tolerance, just rescaled to the
unnormalized normal so it too avoids the `sqrt`:

```
abs(n̂ ⋅ c) < tol   ⟺   abs(n ⋅ c) / ‖n‖ < tol   ⟺   (n ⋅ c)² < tol² ‖n‖²
```

**The exact-arithmetic and symbolic-perturbation fallback is kept.** When `a`
and `b` are nearly equal or nearly antipodal the Float64 cross product has lost
the normal's direction; that case still calls the full `robust_cross_product`,
guarded by exactly the criterion `robust_cross_product` applies internally. To
keep one source of truth for that criterion, the `kMinNorm` expression is lifted
out of `stable_cross_product` into `RobustCrossProduct.min_stable_norm(T)` —
same expression, same association order, same `Float64` value. It constant-folds;
the emitted fast path has no `fdiv` and no `sqrt`.

Measured **36.8 ns → 4.6 ns per call (8.0x)** over 4096 random triples, and the
same over short-edge "clipping-like" triples. Still zero allocations.

Equivalence, old vs new run side by side over **22.2M triples**:

| suite | disagreements |
|---|---|
| 8.0M uniform random unit triples | 0 |
| 2.0M exactly collinear (`c ∈ span(a,b)`) | 0 |
| 200k near-degenerate `a,b` (identical, 1–4 ulp apart, antipodal, near-antipodal; 177,576 slow path) | 0 |
| 8000 exhaustive over axes/poles/diagonals (760 slow path) | 0 |
| 6.0M with `c` placed deliberately *inside* the band, `abs(n̂ ⋅ c) = ρ·tol`, ρ ∈ [0.95, 1.05] | 371,908 |
| 2.0M near-antipodal `a,b` with `c` in the band (325,066 slow path) | 7,470 |

Every disagreement, in every suite, has `abs(n̂ ⋅ c) / tol` between 0.950 and
1.048 — the ULP noise of the dot product itself, which at a tolerance of only
`16 eps` is a few percent of the tolerance. **Zero disagreements were of the form
+1 vs -1**: the two implementations never disagree about which side a point is
on, only about whether a point within ~5% of the tolerance boundary is inside the
"collinear" band. Those are precisely the points the tolerance declares
indeterminate.

### 2. `SphericalCap` `_intersects`: settle on the squared chord, not `atan2`

`_intersects` formed the full central angle with `spherical_distance` — a cross
product, a `sqrt` and an `atan2` — only to compare it against the sum of the two
radii. The squared chord between the centers is 3 subtractions and 3
multiplications, and it brackets that angle tightly enough to settle almost every
pair outright:

```
c ≤ d = 2 asin(c/2) ≤ c (1 + c²/20)      for 0 ≤ c ≤ 1
```

so `c² > r²` rejects and `c²(1 + c²/20)² ≤ r²` accepts, both `sqrt`-free;
`r ≥ π` accepts outright. Only pairs left inside the band pay for
`spherical_distance`, and there the original expression runs unchanged.

The upper bound was verified in BigFloat over 200k values of `c ∈ (0, 1]`:
minimum slack 1.04e-18, i.e. it never crosses. Both cheap branches must clear
the decision by `2^-40` relative — ~1000x the rounding of a squared chord — so
neither can fire on the strength of its own round-off. Whole-sphere (NaN-radius)
caps, which `minimum_bounding_circle` returns for empty input, are unchanged:
every comparison here is false against `NaN`, so such a pair falls through to the
original test and still answers `false`. Same for a NaN center, and for negative
radii.

Measured, 4096 cap pairs each:

| case | before | after |
|---|---:|---:|
| random centers, radii `[0, π)` | 16.5 ns | 7.35 ns (2.25x) |
| random centers, radii ~1e-3 | 16.6 ns | 3.43 ns (4.83x) |
| near centers (sep 1e-2), r ~1e-3 | 12.9 ns | 3.49 ns (3.70x) |
| near centers (sep 1e-3), r ~1e-3 | 12.8 ns | 3.80 ns (3.36x) |
| near centers (sep 1e-3), r ~1e-2 | 12.9 ns | 4.28 ns (3.02x) |

Equivalence over 8.5M pairs: **0 disagreements** for uniform radii in `[0, π)`,
for radii < 0.01, for radii < 1e-6, for identical and exactly-antipodal centers,
and for 2M pairs with `rx + ry` tuned to within 1e-13 relative of the true center
distance.

The two *do* differ once the centers are closer than ~1e-6 rad and `rx + ry` is
within a couple of ULPs of that distance — **and there the old test is the one
that is wrong.** `spherical_distance`'s `cross(p, q)` cancels catastrophically
for nearly equal `p, q`, giving the angle a relative error of order `eps/d`,
while the chord is computed accurately in that regime. Adjudicated against
256-bit arithmetic over 600k such pairs: 111,514 disagreements, old wrong on
111,514 of them, new wrong on 0. So this is a small accuracy improvement as well
as a speedup.

## Why it's worth doing

These came out of a downstream profiling campaign on a CopDEM → IGeo7 conservative
regrid, where `spherical_orient` is called **114,212,848 times per output column**
(it backs `point_on_spherical_arc`, the ring-containment crossing parities, and
the spherical Sutherland–Hodgman clip) and the cap-overlap test was 24.7% of CPU
in the tree walk. Repinning GeometryOps to these two commits measured
**−15% end-to-end, bit-identical output, reproduced twice**, with allocation
unchanged — as expected of two changes that only remove arithmetic.

Measurement records, with the benchmark harnesses and the equivalence suites:

- `DiscreteGlobalGrids.jl/regrid-notes/2026-08-20-perf-ladder.md` (§2.7)
- `DiscreteGlobalGrids.jl/regrid-notes/2026-08-20-more-improvements.md` (§2.2, §2.4)

## Tests

GeometryOps' own suite is unchanged across the pair: **231,527 pass / 0 fail,
per-testset identical** on both sides.

https://claude.ai/code/session_01AC7rbi9eqZypWseMjBGSJ2

### DIFF
--- a/src/utils/UnitSpherical/cap.jl
+++ b/src/utils/UnitSpherical/cap.jl
@@ -103,35 +103,31 @@ const _CAP_CHORD_GUARD = 2.0^-40
 # This is all inexact and thus subject to floating point error
 #=
 Two caps intersect iff the angle `d` between their centers is at most the sum
-`r` of their radii.  Computing `d` costs a `spherical_distance` — a cross
-product, a `sqrt` and an `atan2` — but the *squared chord* between the centers,
-`c² = ‖p - q‖²`, is 3 subtractions and 3 multiplications, and it brackets `d`
+`r` of their radii.  We can shortcut by computing the *squared chord* between the centers,
+`c² = ‖p - q‖²`, which is 3 subtractions and 3 multiplications. It brackets `d`
 tightly enough to settle almost every pair without ever forming `d`:
 
     c  ≤  d = 2 asin(c/2)  ≤  c * (1 + c²/20)        for 0 ≤ c ≤ 1
 
 (the lower bound is `asin(x) ≥ x`; for the upper, `2 asin(c/2) - c` has the
 series `c³/24 + 3c⁵/640 + 15c⁷/21504 + …`, which stays under `c³/20` on `c ≤ 1`
-— verified in BigFloat, minimum slack `1.0e-18` at `c → 0` and `2.6e-3` at
-`c = 1`).  Squaring both sides keeps it `sqrt`-free.  Only pairs that land
-inside the resulting band pay for `spherical_distance`, and there the original
-test runs unchanged.
+, minimum slack `1.0e-18` at `c → 0` and `2.6e-3` at `c = 1`).
+Squaring both sides keeps it `sqrt`-free.
 
-`c ≤ 1` is `d ≤ 60°`; wider pairs skip the accept bound but still get the
-reject one, and `r ≥ π` accepts outright since no two points are further
-than `π` apart.
+This acts as an adaptive filter, similar to how the predicates work.
+Anything which is sufficiently far away gets caught by this, closer things
+go to the old comparison for accuracy.
 
-Every comparison is written so that a NaN radius (the whole-sphere cap that
-`minimum_bounding_circle` returns for empty input) fails it and falls through
-to the original test, which is what `NaN` did before.  Same for negative radii.
+The accept bound requires `c ≤ 1` (`d ≤ 60°`), but the reject bound does not.
+Since no two points are more than `π` apart, `r ≥ π` accepts immediately.
+NaN radii (used for empty-input whole-sphere caps) and negative radii fail the
+filter comparisons and reach the original test, preserving its semantics.
 
-Where this and the old test disagree, the old test is the one that is wrong:
-`spherical_distance`'s `cross(p, q)` cancels catastrophically for nearly equal
-`p, q`, giving `d` a relative error of order `eps/d`, while the chord is
-computed accurately there.  Adjudicated against 256-bit arithmetic over 600k
-pairs whose centers are 1e-12..1 rad apart with `r` within 2 ULP of the true
-distance: 111,514 disagreements, of which the old test got 111,514 wrong and
-this one got 0 wrong.
+For nearly equal centers, `spherical_distance` suffers catastrophic cancellation
+in `cross(p, q)` and relative error of order `eps/d`; the chord remains accurate.
+Against 256-bit arithmetic on 600k pairs 1e-12..1 rad apart, with `r` within
+2 ULP of the true distance, all 111,514 disagreements were old-test errors;
+the chord test had none.
 
 Assumes unit-length centers, as the rest of the cap API does.
 =#
--- a/src/utils/UnitSpherical/cap.jl
+++ b/src/utils/UnitSpherical/cap.jl
@@ -94,10 +94,63 @@ end
 
 # TODO: this returns an approximately antipodal point...
 
+# The margin by which the cheap bounds in `_intersects` must clear the decision
+# before they are trusted: comfortably above the rounding of a squared chord
+# (~4 eps relative) and comfortably below any distance anyone cares about.
+const _CAP_CHORD_GUARD = 2.0^-40
+
 # TODO: exact-predicate intersection
 # This is all inexact and thus subject to floating point error
+#=
+Two caps intersect iff the angle `d` between their centers is at most the sum
+`r` of their radii.  Computing `d` costs a `spherical_distance` — a cross
+product, a `sqrt` and an `atan2` — but the *squared chord* between the centers,
+`c² = ‖p - q‖²`, is 3 subtractions and 3 multiplications, and it brackets `d`
+tightly enough to settle almost every pair without ever forming `d`:
+
+    c  ≤  d = 2 asin(c/2)  ≤  c * (1 + c²/20)        for 0 ≤ c ≤ 1
+
+(the lower bound is `asin(x) ≥ x`; for the upper, `2 asin(c/2) - c` has the
+series `c³/24 + 3c⁵/640 + 15c⁷/21504 + …`, which stays under `c³/20` on `c ≤ 1`
+— verified in BigFloat, minimum slack `1.0e-18` at `c → 0` and `2.6e-3` at
+`c = 1`).  Squaring both sides keeps it `sqrt`-free.  Only pairs that land
+inside the resulting band pay for `spherical_distance`, and there the original
+test runs unchanged.
+
+`c ≤ 1` is `d ≤ 60°`; wider pairs skip the accept bound but still get the
+reject one, and `r ≥ π` accepts outright since no two points are further
+than `π` apart.
+
+Every comparison is written so that a NaN radius (the whole-sphere cap that
+`minimum_bounding_circle` returns for empty input) fails it and falls through
+to the original test, which is what `NaN` did before.  Same for negative radii.
+
+Where this and the old test disagree, the old test is the one that is wrong:
+`spherical_distance`'s `cross(p, q)` cancels catastrophically for nearly equal
+`p, q`, giving `d` a relative error of order `eps/d`, while the chord is
+computed accurately there.  Adjudicated against 256-bit arithmetic over 600k
+pairs whose centers are 1e-12..1 rad apart with `r` within 2 ULP of the true
+distance: 111,514 disagreements, of which the old test got 111,514 wrong and
+this one got 0 wrong.
+
+Assumes unit-length centers, as the rest of the cap API does.
+=#
 function _intersects(x::SphericalCap, y::SphericalCap)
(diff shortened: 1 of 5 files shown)


src/utils/UnitSpherical/robustcrossproduct/RobustCrossProduct.jl
```diff
@@ -68,6 +68,25 @@ const HAS_LONG_DOUBLE = precision(Float64) < precision(BigFloat)
...
+[`robust_cross_product`](@ref) must escalate to higher precision, exact
+arithmetic, or symbolic perturbation.
+
+Callers that want the cheap unnormalized `cross(a - b, a + b)` directly (the
+orientation predicate, say, which only needs the *sign* of a dot product
+against it, and so needs neither the normalization nor a scale-correct norm)
+can use this to decide when the cheap value is trustworthy and when they must
+fall back to `robust_cross_product`.
+
+See the derivation in [`stable_cross_product`](@ref).
```
asinghvi17: WALL OF TEXT please decrease and make this human readable

src/utils/UnitSpherical/cap.jl
```diff
@@ -94,10 +94,63 @@ end
...
+`minimum_bounding_circle` returns for empty input) fails it and falls through
+to the original test, which is what `NaN` did before.  Same for negative radii.
+
+Where this and the old test disagree, the old test is the one that is wrong:
+`spherical_distance`'s `cross(p, q)` cancels catastrophically for nearly equal
+`p, q`, giving `d` a relative error of order `eps/d`, while the chord is
+computed accurately there.  Adjudicated against 256-bit arithmetic over 600k
+pairs whose centers are 1e-12..1 rad apart with `r` within 2 ULP of the true
+distance: 111,514 disagreements, of which the old test got 111,514 wrong and
+this one got 0 wrong.
```
asinghvi17: Holy shit, wall of text.


```suggestion
#=
Two caps intersect iff the angle `d` between their centers is at most the sum
`r` of their radii.  We can shortcut by computing the *squared chord* between the centers,
`c² = ‖p - q‖²`, which is 3 subtractions and 3 multiplications. It brackets `d`
tightly enough to settle almost every pair without ever forming `d`:

    c  ≤  d = 2 asin(c/2)  ≤  c * (1 + c²/20)        for 0 ≤ c ≤ 1

(the lower bound is `asin(x) ≥ x`; for the upper, `2 asin(c/2) - c` has the
series `c³/24 + 3c⁵/640 + 15c⁷/21504 + …`, which stays under `c³/20` on `c ≤ 1`
, minimum slack `1.0e-18` at `c → 0` and `2.6e-3` at `c = 1`).  
Squaring both sides keeps it `sqrt`-free.

This acts as an adaptive filter, similar to how the predicates work.
Anything which is sufficiently far away gets caught by this, closer things
go to the old comparison for accuracy.
```

Below here you will have to edit the text to a similar style.

`c ≤ 1` is `d ≤ 60°`; wider pairs skip the accept bound but still get the
reject one, and `r ≥ π` accepts outright since no two points are further
than `π` apart.

Every comparison is written so that a NaN radius (the whole-sphere cap that
`minimum_bounding_circle` returns for empty input) fails it and falls through
to the original test, which is what `NaN` did before.  Same for negative radii.

Where this and the old test disagree, the old test is the one that is wrong:
`spherical_distance`'s `cross(p, q)` cancels catastrophically for nearly equal
`p, q`, giving `d` a relative error of order `eps/d`, while the chord is
computed accurately there.  Adjudicated against 256-bit arithmetic over 600k
pairs whose centers are 1e-12..1 rad apart with `r` within 2 ULP of the true
distance: 111,514 disagreements, of which the old test got 111,514 wrong and
this one got 0 wrong.
```

File: src/utils/UnitSpherical/predicates.jl [hunk 3]
```diff
@@ -26,10 +26,60 @@ spherical_orient(a, b, c)
...
+a few percent of `tol`, since `tol` is only `16 * eps`).  Points that close to
+the boundary are exactly the ones the tolerance declares indeterminate; no
+input can flip between `+1` and `-1`, only between `0` and a sign.
+
+When the cheap cross product is *not* trustworthy — `a` and `b` nearly equal or
+nearly antipodal, i.e. `‖n‖ < min_stable_norm(T)` — this falls back to the full
+`robust_cross_product`, including its extended/exact-precision and symbolic
+perturbation paths.  That fallback is unreachable from the Sutherland-Hodgman
+clipping path (which skips `edge_start == edge_end` edges), but this is a
+general-purpose predicate with other callers, so it stays.
```
asinghvi17: WALL OF TEXT - same treatment as the last one.

src/utils/UnitSpherical/robustcrossproduct/RobustCrossProduct.jl
```diff
@@ -68,6 +68,15 @@ const HAS_LONG_DOUBLE = precision(Float64) < precision(BigFloat)
...
 const EXACT_CROSS_PROD_ERROR = DBL_ERR
 
+"""
+    min_stable_norm(::Type{T})
+
+Return the smallest stable-cross-product norm that meets
+`ROBUST_CROSS_PROD_ERROR` in precision `T`.
+"""
+min_stable_norm(::Type{T}) where {T} =
+    (32 * SQRT3 * DBL_ERR) / (ROBUST_CROSS_PROD_ERROR / (eps(float(T)) / 2) - (1 + 2 * SQRT3))
```
asinghvi17, line 80: Where does this expression come from?

asinghvi17: Oh I see from the diff below.  Just point to the permalink in s2 then.

asinghvi17:
Approved, good job on removing the walls of text.  Just address this comment and I think we are good to merge.

asinghvi17:
Rebased the whole stack onto current `main` (`3b1878db0`, #472). No new work — the SHA change is the rebase only.

| PR | branch | before | after |
|----|--------|--------|-------|
| #476 | `claude/perf-ladder-predicates` | `1f38e099a` | `2c0e093f3` |
| #479 | `claude/spherical-cap-primitives` | `697c7cc81` | `eeec12d50` |
| #485 | `as/spherical-lightweight-predicates` | `60e83c3dc` | `3563b9069` |
| #486 | `as/spherical-fh-clipping` | `8af92af81` | `35996798b` |

Zero conflicts. Two things are worth calling out, because they change content *below* this PR:

1. **#479/#485/#486 now carry this PR's review-response commits.** They were branched at `3706b6cca`, so they never had the last five commits here. Those are now in, which is why `cap.jl`'s filter comment reads in the shortened form on the lower branches, and why `f188da366` (the `kMinNorm` permalink) is present throughout.
2. **The `skip = VERSION < v"1.12"` guard is on all four branches for the first time.** `f3b73846f` had reverted it; `1eb7cf700` restored it here, but the lower branches predated both. Its absence is what made Julia 1.11 red on #479/#485/#486 — that should now clear.

Verification before pushing: each rebased tip's tree is byte-identical to independently merging `main` + this branch into the corresponding old tip (`git diff` empty on all four), the chain is linear, and the top of the stack loads and smoke-tests clean. Pre-rebase heads are tagged locally as `backup/prerebase/*`.

Issue #475, opened 2026-08-19
Leftover `Any` fields and needless mutable structs in the OverlayNG / RelateNG ports


asinghvi17:
Sweep for the two defects fixed in 2825c4766 (`_OverlayInput`): `::Any` fields where a type parameter is derivable, and `mutable struct`s that don't need to be mutable. That commit is the worked example — type params for the eager fields, `Base.RefValue` for the genuinely lazy ones.

Confirmed with `fieldtypes`/`isconcretetype`/`Base.infer_return_type`, not grep. Impact figures are microbenchmarks of the isolated call (boxed vs typed), so they bound the per-call delta rather than measuring a whole-op speedup.

## Hot-path dynamic dispatch

**1. `_OverlayEdgeRing.locator::Any` — `src/methods/clipping/overlayng/maximal_edge_ring.jl:56`**

The same field, with the same `# Union{Nothing, IndexedPointInAreaLocator}, lazy` comment, as the `_OverlayInput.loc_a` that 2825c4766 fixed, and the struct's only non-concrete field. `_ring_locate` (`:971`) infers `Any`, so `locate(ring.locator, p)` is a dynamic dispatch, read per hole vertex by `_is_point_in_or_out` (`:995`). Boxed vs typed `locate`: +9–10 ns, +13–18% at ring sizes 8/40/200. Free-hole placement is documented as rare (405 placements over the robust corpus), so the aggregate is small.

`_PolyBuilderCtx{M,P,E,T,NB}` already carries `M` and `E`, and the locator type is `IndexedPointInAreaLocator{M,E}` — either add those params to `_OverlayEdgeRing`, or hold the locators on the ctx in a `Vector{Union{Nothing, IndexedPointInAreaLocator{M,E}}}` keyed by ring id, as `RelatePointLocator.poly_locator` already does. `_edge_ring_type(T)` would become `_edge_ring_type(m, exact, T)`.

**2. `RelatePointLocator.lines::Vector{Any}` / `polygons::Vector{Any}` — `src/methods/geom_relations/relateng/point_locator.jl:261-262`**

[...]

PR #478, opened 2026-08-21
Typed OverlayNG and RelateNG storage

asinghvi17:
## Summary
- replace boxed OverlayNG and RelateNG hot-path fields with concrete parameterized storage
- make ring, builder-context, and node-section structs immutable while retaining lazy state in refs and mutable containers
- derive point-in-area index storage from the manifold and narrow homogeneous locator element vectors
- require GeoInterface 1.6.1 for total empty-geometry point and ring counts
- add regression coverage for field concreteness, inference, immutability, and heterogeneous collections

## Testing
- `julia --project=. -e "using Pkg; Pkg.test()"`
- focused OverlayNG engine and points suites
- focused RelateNG locator, indexed point-in-area, node topology, and geometry suites

Stacked on #473. Addresses #475.

Written with Codex


asinghvi17:
I (the human) generally have no objections to this

src/methods/geom_relations/relateng/point_locator.jl
```diff
@@ -278,6 +279,11 @@ function RelatePointLocator(m::Manifold, geom; exact,
     lines = Any[]
     polygons = Any[]
     _extract_elements!(m, points, lines, polygons, geom)
+    # Narrow the overwhelmingly common homogeneous case after the recursive
+    # traversal; `map` computes a common element type from the values and keeps
+    # heterogeneous collections widened as needed.
```
asinghvi17, line 284: Comment too long.  Prefer identity.(lines) over map here.  Also, any way we can remove the Any type from the lines and polygons?  Maybe that is work for later.

asinghvi17:
Looks good to me

PR #479, opened 2026-08-22
Publish spherical cap extent primitives

asinghvi17:
Stacked on #476; the diff here is only the cap-primitives work (plus the merge that reconciled the shared predicate commits).

Makes `UnitSpherical.SphericalCap` a public extent primitive (src/utils/UnitSpherical/cap.jl): the cap type and its intersection/containment predicates get a documented public surface so downstream tree searches (ConservativeRegridding's dual DFS) can use caps as node extents generically instead of through private APIs. Consumed downstream by ConservativeRegridding's extent-generic search (see the `claude/regrid-prereqs` PR there).

https://claude.ai/code/session_01AC7rbi9eqZypWseMjBGSJ2

### DIFF
--- a/src/utils/UnitSpherical/cap.jl
+++ b/src/utils/UnitSpherical/cap.jl
@@ -7,6 +7,10 @@ CollapsedDocStrings = true
 
 ```@docs; canonical=false
 SphericalCap
+cap_contains
+merge_caps
+dilate_cap
+cap_xyz_extent
 circumcenter_on_unit_sphere
 ```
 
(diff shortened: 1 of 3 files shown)


src/utils/UnitSpherical/cap.jl
```diff
@@ -151,17 +156,45 @@ end
 
 _disjoint(x::SphericalCap, y::SphericalCap) = !_intersects(x, y)
 
-function _contains(big::SphericalCap, small::SphericalCap)
+"""
+    Extents.contains(big::SphericalCap, small::SphericalCap; strict=false)
+    Extents.within(small::SphericalCap, big::SphericalCap; strict=false)
```
asinghvi17: Not in the same docstring for another function.  And generally this should be the signature and a one liner.  The meaning is explained in the main Extents.jl docstring.

src/utils/UnitSpherical/cap.jl
```diff
@@ -151,17 +156,45 @@ end
...
 
+# ## Cap intersection
+
+"""
+    Extents.intersects(x::SphericalCap, y::SphericalCap)
+
+Whether the two closed caps intersect. Tangency counts as intersection. The
+comparison delegates to the cap predicate's guarded squared-chord test, which
+retains precision for nearly coincident centres and falls back to angular
+distance only inside its uncertainty band.
```
asinghvi17: Again too long.  This should be a one liner.

src/utils/UnitSpherical/cap.jl
```diff
@@ -190,22 +223,82 @@ end
 Extents.intersects(ext::Extents.Extent{(:X, :Y, :Z)}, cap::SphericalCap) =
     Extents.intersects(cap, ext)
 
-#Comment by asinghvi: this could be transformed to GO.union
-function _merge(x::SphericalCap, y::SphericalCap)
+"""
+    Extents.extent(cap::SphericalCap) -> Extents.Extent{(:X, :Y, :Z)}
```
asinghvi17: This feels a bit dangerous.  How about some form of convert method?  I could imagine something calling extent on arbitrary input just to sanitize.  Especially in user code.

But if you feel it's safe up to you.

src/utils/UnitSpherical/cap.jl
```diff
@@ -190,22 +223,82 @@ end
...
+    return Extents.Extent(X = bounds[1], Y = bounds[2], Z = bounds[3])
+end
+
+"""
+    Extents.grow(cap::SphericalCap, factor::Real) -> SphericalCap
+
+Grow the cap's angular diameter by `factor` on each side, matching the scalar
+`Extents.grow` convention. Positive growth is rounded outward and capped at
+`π`; zero returns `cap` unchanged. A factor of `0.5` therefore doubles the
+radius. Negative factors shrink it, provided the resulting radius is valid.
```
asinghvi17: Too verbose

src/utils/UnitSpherical/cap.jl
```diff
@@ -190,22 +223,82 @@ end
...
+end
+
+"""
+    Extents.union(x::SphericalCap, y::SphericalCap; strict=false) -> SphericalCap
+
+Return a cap containing both `x` and `y`. If either input already contains the
+other it is returned unchanged. Otherwise the centre lies between the input
+centres and the radius is rounded outward to preserve coverage. `strict` is
+accepted for consistency with the Extents API and has no effect because both
+inputs use the same domain.
```
asinghvi17:
```suggestion
Return a cap containing both `x` and `y`. 

If either input already contains the other it is returned unchanged. 
Otherwise, the centre lies between the input caps' centres, and 
the radius is rounded outward to preserve coverage. 

`strict` is accepted for consistency with the Extents API 
and has no effect because both inputs use the same domain.
```

asinghvi17 (commented):
Generally ok but some comments.

asinghvi17:
LGTM but do see why ci fails

PR #493, opened 2026-09-09
Remove boxed ring length from spherical extents

asinghvi17:
Spherical region extents reassigned the ring length before capturing it in a reduction closure, producing `Core.Box` and an `Any` return. Assign the effective length once so the existing extent calculation infers concretely, preserving edge order, closure handling, and conservative region bounds.

On warmed 64- and 1,024-vertex fixtures, extent time fell from 14.50 to 1.42 µs and 169.67 to 24.63 µs respectively; both measured calls now allocate zero bytes. These are isolated extent measurements on Julia 1.12.7.

Validation: 927 extent assertions, including new inference coverage for implicit/explicit closure, the public extent entry point, and interaction bounds. The complete stack was also checked against the spherical predicate, preparation, and OverlayNG suites.

Stack (merge in this order):

2. #493 — Spherical extent boxing (this PR)
3. #494 — Exact spherical crossings
4. #495 — Concrete optional ring locators

Depends on #497; this PR targets that branch so its diff contains only this layer.


asinghvi17:
LGTM

PR #494, opened 2026-09-09
Reuse exact orientation signs for spherical crossings

asinghvi17:
The general spherical ring scan recomputed proper crossings with `Rational{BigInt}` cross/dot arithmetic even after calculating the four exact orientation signs. Reuse the same exact sign decision already used by intersection classification, and let the parity walkers pass their existing signs through an internal callback hook.

Sign reuse is restricted to the recognized exact orientation and crossing callbacks. Custom callbacks, the default tolerance-based callback, approximate arithmetic, and the incidence/degeneracy branches keep their existing behavior. Canonicalize raw `exact=true` so it benefits as well.

Isolated cached ring scans improved from 34.79 to 1.50 µs for 64 vertices and from 55.00 to 23.42 µs for 1,024 vertices. Together with the preceding extent fix, the Natural Earth/HEALPix L8 intersects rasterization improved from 1.54 s / 1.77 GB to 0.63 s / 0.84 GB; center improved from 495 to 430 ms. All four L6/L8 center/intersects full-label checksums matched baseline. These are warm paired measurements; the end-to-end gains belong to the two fixes together.

Validation: 4,161 assertions in the new testset, including an independent rational oracle, dyadic/ULP perturbations, non-unit coordinates, antipodal alternatives, reversals/scaling, both parity walkers, raw Bool exactness, and callback compatibility. The five original targeted files plus new coverage passed 61,783 assertions before stacking; combined-stack coverage was rerun before publication.

Stack (merge in this order):

2. #493 — Spherical extent boxing
3. #494 — Exact spherical crossings (this PR)
4. #495 — Concrete optional ring locators

Depends on #493; this PR targets that branch so its diff contains only this layer.


asinghvi17:
LGTM

PR #495, opened 2026-09-09
Use concrete optional storage for OverlayNG ring locators

asinghvi17:
OverlayNG's lazy ring locator stored either `nothing` or a known internal locator in an `Any` field. Store `Union{Nothing,L}` with concrete `L = IndexedPointInAreaLocator{M,E}`, propagate that type through the builder context, and narrow a local cache value before lookup. Lazy construction and existing mutability are preserved.

The isolated spike measured cached lookups 20–25% faster with 32 fewer allocated bytes per call. Representative full overlays were effectively unchanged. Three alternating fresh-process pairs recorded first-call and compilation counters separately and found no detected compilation regression; the small differences were within run variance. These figures predate stacking and do not measure the combined stack.

Validation: all 813 assertions in the OverlayNG engine file passed, including 44 new assertions covering both manifolds, exactness settings, lazy-cache states, inference, and free-hole placement. Combined-stack coverage was rerun before publication.

Related: #478 proposes broader typed storage and immutability changes in this area. This PR isolates the optional locator union; it intentionally overlaps part of that proposal and should be reconciled with it when merging.

Stack (merge in this order):

2. #493 — Spherical extent boxing
3. #494 — Exact spherical crossings
4. #495 — Concrete optional ring locators (this PR)

Depends on #494; this PR targets that branch so its diff contains only this layer.


asinghvi17:
LGTM

PR #500, opened 2026-09-11
Complete manifold propagation in Foster–Hormann clipping

asinghvi17:
Foster–Hormann clipping accepted a spherical manifold but still used planar predicates in hole handling, multipolygon corrections, and point-intersection filtering. This completes those paths, fixes the deferred Gambia–Senegal enclave error, and reduces repeated longitude/latitude conversion. Builds on #490; no existing PR is merged.

- Pass the manifold and numeric type through hole predicates, recursive clipping, and multipolygon corrections. Default corrections inherit the caller's algorithm/T; explicit corrections retain their settings. Preserve outer correction CRS and support empty multipolygon results.
- Use one typed spherical vertex conversion for both node lists, convert computed crossings to the requested type, and preserve the first input's representation—including mixed-input and bit-exact passthrough cases. This fixes the Float32 `PolyNode` conversion failure.
- Remove the invalid planar envelope rejection from spherical `intersection_points`. Explicit planar tree accelerators now reject spherical input with an actionable error; `NestedLoop` and `AutoAccelerator` are supported.
- Preserve Cartesian coordinates in shared relation processors. Make `cut` preserve representation and pair crossings in line order, with both line endpoints outside the polygon.
- Fix the planar shared-boundary enclave classification by probing an off-boundary edge when all vertices are shared. The compact triangle/notched-square regression covers winding, rotation, operand order, and cache reuse; the original Natural Earth Gambia–Senegal intersection is now zero in both orders.
- Materialize longitude/latitude inner edges once per clip and reuse that buffer in `FosterHormannCache`. Already-Cartesian input keeps its allocation-free iterator.

Validation: **8,329 assertions** across the clipping, spherical FH, spherical relation, correction, intersection-area, intersection-point, cut, and new regression suites passed on Julia 1.13. The new tests include **18 independent S2 hole comparisons**. All **four doctest blocks** from the changed clipping files pass in a focused Documenter run.

Matched warm `intersection_area` benchmarks, 10,000 samples with baseline/current measured sequentially (microseconds):

| Input / vertices | Cached baseline | Cached current |
| --- | ---: | ---: |
| lon/lat / 6 | 6.250 | 5.000 |
| lon/lat / 33 | 87.708 | 45.791 |
| Cartesian / 6 | 2.333 | 2.333 |
| Cartesian / 33 | 33.083 | 33.000 |

Warm cached calls remain **96 B / 4 allocations** in all four cases. Uncached longitude/latitude calls add two allocations for the converted edge vector: 368 B at 6 vertices and 1,664 B at 33 vertices. Two fresh-process measurements put first longitude/latitude-call compilation at **2.56–2.64 s before vs 3.02–3.34 s after**; the runtime improvement has a real compilation cost. Benchmark scripts and development logs are not committed.

A remaining limitation is now documented on `FosterHormannClipping`: exact predicates classify the input arcs, but computed crossings are rounded. Arbitrary repeated clipping of an earlier Cartesian result against the same boundary can still produce inconsistent representable topology and `TracingError`. Default multipolygon correction now also normalizes the subtracted components, avoiding the reproduced duplicate-component case at an additional correction cost; `fix_multipoly=nothing` still opts out. Fully robust arbitrary chaining requires intersection provenance/exact constructions or a separately designed snapping policy, not an ad hoc tolerance. This PR does not claim new support for intentionally unsupported algorithms/manifolds.


asinghvi17:
Folded into #486, including the shared-boundary enclave fix, complete manifold/type propagation, and the Float32 exact-orientation prerequisite. #487 and #490 were restacked above the combined #486 layer, and the resulting #490 tree is equivalent to this PR plus the measured Foster–Hormann precompile workload. No PRs were merged.

PR #503, opened 2026-09-16
Convert coordinates to T in spherical point and ring extents

evetion:
Float32 input reached the exact predicates in the ring orientation check and threw; the point extent also ignored T.


asinghvi17 (approved):
thanks!

PR #456, opened 2026-08-04
Use crstrait to correctly derive area.

evetion:
This defaults area to use AutoManifold, which in turn results in

Without Proj
projected/unknown => planar
geographic => spherical

With Proj
unknown => Proj check
geographic => geodesic

Also includes a geodesic area fix

PR #504, opened 2026-09-16
Select the manifold from the CRS in `perimeter`, `distance` and `segmentize`

evetion:
Follow-up to #456, which made `area` CRS-aware. This applies the same `AutoManifold` default to the other measurement functions.

## Changes

- The CRS-to-manifold resolution moves out of `area` into a shared internal `_auto_manifold` (`src/utils/utils.jl`), which the Proj extension extends. It returns the manifold and the degrees-per-unit scales of the longitude/latitude axes, so non-degree CRSs such as EPSG:4807 (grads) work.
- `perimeter`, `distance` and `segmentize` default to `AutoManifold()`:
  - With Proj loaded, recognized geographic CRSs use `Geodesic` on the CRS ellipsoid, in metres.
  - Without Proj, geographic CRS traits use `Spherical`.
  - Projected, unknown and CRS-less inputs use `Planar`, so their results are unchanged.
- `distance(a, b)` takes the CRS from whichever input has one. If both have a CRS and they differ, it throws an `ArgumentError`.
- `distance` supports lines, rings and polygons on `Spherical` and `Geodesic`. Before, only point to point worked, and only on `Spherical`.
  - Each manifold provides a point-point and a point-segment kernel; the curve and polygon logic is shared.
  - Spherical point to arc is closed-form.
  - Geodesic point to segment finds the perpendicular foot with a bracketed Illinois root find on the Proj geodesic. It matches brute-force sampling to about 1e-16 relative.
  - Geodesic point-in-polygon uses the spherical test.
- Geodesic `perimeter` creates its Proj objects per curve, so it is safe with `threaded = true`.
- Geodesic `segmentize` uses `_flattening`, which handles `inv_flattening = 0` (a sphere).
- The undocumented trait-dispatched `distance(trait1, g1, trait2, g2)` methods without a manifold are removed; the manifold versions cover them.
- `perimeter` gets a docstring, and the CRS and manifold explanation pages are updated.

## Behaviour change

Calls without a manifold on geometries with a geographic CRS now return metres instead of degrees, as `area` already does. For `segmentize`, `max_distance` is then also in metres.

## Tests

- `AutoManifold` tests for all three functions: EPSG:4326, EPSG:2263, EPSG:4807, CRS-less inputs, threaded runs, and mismatched CRSs.
- Spherical and geodesic `distance` tests against closed-form and sampled references.
- The no-Proj subprocess test is renamed to `auto_manifold_no_proj.jl` and covers the `Spherical` path through a geometry type whose `crstrait` is geographic.

The full test suite passes locally.


ext/GeometryOpsProjExt/distance.jl
```diff
@@ -0,0 +1,63 @@
+import GeometryOps: _distance_point, _distance_segment
+
+function GeometryOps._distance_auto(m::Geodesic, geom1, geom2, ::Type{T}, (longitude_scale, latitude_scale); kwargs...) where T
+    longitude_scale == latitude_scale == 1 && return GeometryOps.distance(m, geom1, geom2, T; kwargs...)
```
evetion, line 4: @asinghvi17 so this comes from the previous area PR (#456), for CRSs that have non-degree units (grads, radians).

